PR #3, opened 2025-04-25
🐛 Handle PWA UI for iOS

nwingt:


nuxt.config.ts
```diff
@@ -80,6 +80,8 @@ export default defineNuxtConfig({
           '\'self\'',
           'blob:',
         ],
+        // NOTE: Resolve Safari force HTTPS in development
+        'upgrade-insecure-requests': process.env.NODE_ENV === 'development' ? false : true,
```
williamchong, line 84: Who sets `NODE_ENV`? Would `!== 'production'` be more robust?

nwingt: I think this should be always turned on for other cases
Ref: https://nuxt-security.vercel.app/advanced/faq#running-app-with-host-flag

PR #4, opened 2025-04-25
[TAS-3650] 👷 Add apphosting env

williamchong:


apphosting.mainnet.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - variable: MAGIC_LINK_CUSTOM_LOGO_URL
+    value: https://liker.land/logo.png
+  - variable: MAGIC_LINK_RPC_URL
+    value: https://mainnet.optimism.io
+  - variable: NUXT_SESSION_PASSWORD
+    secret: V3_NUXT_SESSION_PASSWORD
+  - variable: PDF_VIEWER_ORIGIN
+    value: https://likecoin.github.io
+  - variable: PDF_VIEWER_PATH
+    value: /pdf.js/web/viewer.html
```
nwingt: Should we use legacy?

PR #8, opened 2025-05-15
[TAS-3657] ✨ Support EVM NFT Class for shelf & reader

nwingt:


apphosting.mainnet.yaml
```diff
@@ -23,6 +23,8 @@ env:
     value: https://api.like.co
   - variable: LIKECOIN_CHAIN_API_ENDPOINT
     value: https://mainnet-node.like.co
+  - variable: LIKECOIN_EVM_CHAIN_API_ENDPOINT
+    value: https://likenft-indexer.pandawork.com/api
```
williamchong:
```suggestion
    value:
```
Better leave empty until we have mainnet indexer

i18n/locales/zh-Hant.json
```diff
@@ -1,7 +1,10 @@
 {
+  "account_page_update_settings_error": "未能更新帳戶設定。",
   "app_header_shelf": "書架",
   "app_header_store": "書店",
   "bookshelf_more_menu_title": "更多選項",
+  "bookshelf_no_items": "你的書架為空",
```
williamchong:
```suggestion
  "bookshelf_no_items": "你的書架沒有書",
```
or just simply `你的書架是空的`

server/api/account/settings.post.ts
```diff
@@ -0,0 +1,10 @@
+export default defineEventHandler(async (event) => {
+  const session = await requireUserSession(event)
+  const body: { isEVMModeActive?: boolean } = await readBody(event)
+  await setUserSession(event, {
+    user: {
+      ...session.user,
+      isEVMModeActive: body.isEVMModeActive || session.user.isEVMModeActive || false,
```
williamchong: Should these be `??` instead of `||`?  Do we expect `session.user.isEVMModeActive` to override `body.isEVMModeActive`?

composables/use-book-info.ts
```diff
@@ -3,63 +3,62 @@ export default function ({ nftClassId = '' }: { nftClassId?: string } = {}) {
   const metadataStore = useMetadataStore()
   const bookstoreStore = useBookstoreStore()
 
+  const isEVM = computed(() => checkIsEVMAddress(nftClassId))
```
williamchong, line 6: Can we use 2 different composable for evm vs non-evm, then merge them with 1 if else in this composable, and export as unified interface?

e.g.
```
const isEVM = computed(() => checkIsEVMAddress(nftClassId))
const implementation = isEVM ? useLegacyNFTBookInfo(nftClassId) : useEVMBookInfo(nftClassId)
return {
  isEVM,
  ...implementation,
} as ????
```

stores/nft.ts
```diff
@@ -24,14 +30,36 @@ export const useNFTStore = defineStore('nft', () => {
...
-  function addNFTClass(nftClass: NFTClass) {
-    nftClassByIdMap.value[nftClass.id] = nftClass
+  function addLegacyNFTClass(nftClass: LegacyNFTClass) {
+    legacyNFTClassByIdMap.value[nftClass.id] = nftClass
   }
 
-  function addNFTClasses(nftClasses: NFTClass[]) {
+  function addLegacyNFTClasses(nftClasses: LegacyNFTClass[]) {
     const nftClassIds: string[] = []
     for (const nftClass of nftClasses) {
```
williamchong, line 39: Not sure if it is better to just use `addNFTClass` and detect `isEVM` inside `stores/nft`, that way user don't need to know if they are using Legacy or EVM NFT store.
Is splitting these action/getter is meaningful for components? We can keep the state splitted

server/api/account/settings.post.ts
nwingt: Updated

composables/use-book-info.ts
nwingt: Split into 2 composables

https://github.com/likecoin/liker-land-v3/pull/8/commits/0816b765f79e7ba47ccde565fc9759a03131c4b7

stores/nft.ts
nwingt: I tried to use the same method to support different types of NFTClass, but it turned out to be a TypeScript nightmare

composables/use-evm-book-info.ts
```diff
@@ -0,0 +1,51 @@
...
+    const datePublished = contentMetadata.value?.datePublished
+    return datePublished ? new Date(datePublished) : undefined
+  })
+
+  const releasedDate = computed(() => {
+    if (publishedDate.value) return publishedDate.value
+    return new Date(contentMetadata.value?.recordTimestamp || Date.now())
+  })
+
+  const externalURL = computed(() => contentMetadata.value?.external_url)
```
williamchong:
```suggestion
  const externalURL = computed(() => contentMetadata.value?.external_link)
```

As per https://docs.opensea.io/docs/contract-level-metadata

williamchong (approved):

PR #14, opened 2025-05-24
🐛 Fix unable to fetch bookshelf items

nwingt:


pages/shelf.vue
```diff
@@ -107,7 +107,7 @@ const shouldLoadMore = useElementVisibility(infiniteScrollDetectorElement)
 
 onMounted(async () => {
   if (hasLoggedIn.value) {
-    await bookshelfStore.fetchItems()
+    await bookshelfStore.fetchItems({ isRefresh: true })
```
williamchong: How about we dont block the initial fetch even without isRefresh:true ? ie isRefresh is only used for clearing existing items and fetch

nwingt: Updated

williamchong (approved):

PR #22, opened 2025-05-30
💄 Refactor UI before adding purchase flow

nwingt:


components/BookLoadingScreen.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+  <div class="flex flex-col justify-center items-center py-[56px]">
+    <NuxtLink :to="$localeRoute({ name: 'shelf' })">
+      <AppLogo
+        class="text-green-500 shrink-0"
+        :height="24"
+      />
+    </NuxtLink>
+
+    <BookCover
```
williamchong, line 10: Maybe we can do `<model-viewer>` here again

williamchong (approved):

PR #26, opened 2025-05-31
✨ Add reader page transition & update translations

nwingt:


i18n/locales/en.json
```diff
@@ -1,4 +1,16 @@
 {
+  "account_page_account_title": "User Account",
+  "account_page_contact_support": "Contact Us",
+  "account_page_cosmos_wallet": "Cosmos Wallet Address",
```
williamchong: I think it is `LikeCoin Chain` instead of `Cosmos`

i18n/locales/zh-Hant.json
```diff
@@ -1,4 +1,16 @@
 {
+  "account_page_account_title": "個人賬戶",
+  "account_page_contact_support": "聯絡我們",
+  "account_page_cosmos_wallet": "Cosmos 地址",
```
williamchong: I think it is `LikeCoin Chain` instead of `Cosmos`

PR #27, opened 2025-06-01
⬆️ Update Magic Link Wagmi connector

nwingt:
Based on #26


stores/account.ts
```diff
@@ -130,6 +148,27 @@ export const useAccountStore = defineStore('account', () => {
     await refreshSession()
   }
 
+  async function exportPrivateKey() {
+    // NOTE: This function is only for login with Magic Link
+    if (user.value?.loginMethod !== 'magic') return
```
williamchong: Better throw error here

stores/account.ts
```diff
@@ -65,6 +66,19 @@ export const useAccountStore = defineStore('account', () => {
       if (status.value !== 'success') {
         await connectAsync({ connector })
       }
+      let email: string | undefined
+      if ('magic' in connector) {
```
williamchong: Why not just check `connector.id`? Also should we check `status.value` first?

nwingt: maybe check both

williamchong (approved):

PR #28, opened 2025-06-01
🐛 Fix Liker ID info fetching

nwingt:


server/api/login.post.ts
```diff
@@ -77,20 +77,24 @@ export default defineEventHandler(async (event) => {
...
-      }>(`${config.public.likeCoinAPIEndpoint}/users/addr/${body.walletAddress}/min`)
-      likerId = userInfoRes.user
-      displayName = userInfoRes.displayName
-      avatar = userInfoRes.avatar
-      description = userInfoRes.description
-    }
-    catch {
-      console.warn('Failed to fetch user info for wallet')
+    // NOTE: Although the API supports querying user public info by EVM wallet address,
+    // it is not guaranteed that the EVM wallet address is set in the user db.
```
williamchong, line 81: They are supposed to be set, and will be set in user migration API.
I probably forgot to set them for the v3 preview user list, since they were related to liker id at that moment
We should just set them in db

PR #29, opened 2025-06-02
[TAS-3835] 🐛 Fix unable to load book when missing nft_id

nwingt:


composables/use-reader.ts
```diff
@@ -2,12 +2,20 @@ export default function () {
...
   const bookInfo = useBookInfo({ nftClassId: nftClassId.value })
 
   const ownerFirstNFTId = computed(() => {
-    return nftStore.getNFTClassFirstNFTIdByNFTClassIdAndOwnerWalletAddress(nftClassId.value, (bookInfo.isEVM.value ? user.value?.evmWallet : user.value?.likeWallet) || '')
+    let nftId: string | undefined
+    const ownerWallet = bookInfo.isEVM.value ? user.value?.evmWallet : user.value?.likeWallet
+    nftId = nftStore.getNFTClassFirstNFTIdByNFTClassIdAndOwnerWalletAddress(nftClassId.value, ownerWallet || '')
+    if (!nftId) {
+      const nfts = bookshelfStore.getNFTsByNFTClassId(nftClassId.value)
+      if (nfts.length) nftId = nfts[0].token_id
```
williamchong: What is the context of this? Can we just not set any `nftId`?

williamchong: in book function we only requires `nft_id` if it is `mustClaimToView || isCustomMessagePageEnabled`, in either case user should own a nft for it. Or is `isCustomMessagePageEnabled` default on now?

williamchong: Do we still need this PR after https://github.com/likecoin/likecoin-cloud-functions/pull/206 is merged? @nwingt

nwingt: We still need this as we depends on getting nftId from bookshelf (/tokens API)

composables/use-reader.ts
```diff
@@ -2,12 +2,20 @@ export default function () {
...
   const bookInfo = useBookInfo({ nftClassId: nftClassId.value })
 
   const ownerFirstNFTId = computed(() => {
-    return nftStore.getNFTClassFirstNFTIdByNFTClassIdAndOwnerWalletAddress(nftClassId.value, (bookInfo.isEVM.value ? user.value?.evmWallet : user.value?.likeWallet) || '')
+    // Find the first NFT Id from bookshelf
+    let nftId = bookshelfStore.getNFTsByNFTClassId(nftClassId.value).map(nft => nft.token_id)[0]
+    if (!nftId) {
+      // Find the first NFT Id owned by the user for the given NFT class
+      const ownerWallet = bookInfo.isEVM.value ? user.value?.evmWallet : user.value?.likeWallet
+      nftId = nftStore.getNFTIdsByNFTClassIdAndOwnerWalletAddress(nftClassId.value, ownerWallet || '')[0]
```
williamchong, line 16: These 2 stores actually use the same data source (indexer), should merge them in the future

williamchong (approved):

PR #39, opened 2025-06-05
🚸 Add nft_id in reader URL if not set

williamchong:


composables/use-reader.ts
```diff
@@ -1,13 +1,20 @@
...
   const nftClassId = computed(() => getRouteQuery('nft_class_id'))
   const bookInfo = useBookInfo({ nftClassId: nftClassId.value })
 
-  const nftId = computed(() => {
-    const id = getRouteQuery('nft_id')
-    return id !== undefined && id !== '' ? id : bookInfo.userOwnedNFTIds.value[0]
-  })
+  const nftId = computed(() => getRouteQuery('nft_id'))
+  if (!nftId.value) {
+    navigateTo({
```
nwingt: I dont think navigate in composable during setup is a good practice

nwingt: should move this to reader page instead?

nwingt: on 2nd thought
actually we dont need nftId in the qs as we got the `nftId` in `useBookInfo()` of `reader/*` page
in fact we should provide the nftId actively in `BookshelfItem.vue` from the `bookshelfStore` `items`

williamchong: `nft_id` would be still useful if user refresh/come back from bookmark/history, updated logic to be in page

pages/reader.vue
```diff
@@ -22,7 +22,9 @@ const getRouteBaseName = useRouteBaseName()
 const { t: $t } = useI18n()
 
 const nftClassId = computed(() => getRouteQuery('nft_class_id'))
+const nftId = computed(() => getRouteQuery('nft_id'))
 const bookInfo = useBookInfo({ nftClassId: nftClassId.value })
+const { nftId: readerNftId } = useReader()
```
nwingt: No need to add `useReader()`, just refactor `bookInfo.userOwnedNFTIds.value[0]` (e.g. `bookInfo.userFirstOwnedNFTId`) in `useBookInfo()` and use that instead

https://github.com/likecoin/3ook-com/blob/8f701bdf3496ae8b8fd64a8c58f208bb41683030/composables/use-reader.ts#L7-L10

nwingt (approved):

PR #41, opened 2025-06-06
[TAS-3854]💄 Update edition description text style

AuroraHuang22:
before:
截圖 2025-06-06 16 39 59

after:
截圖 2025-06-06 16 39 51

### DIFF
--- a/pages/store/[id].vue
+++ b/pages/store/[id].vue
@@ -157,7 +157,7 @@
                       v-else
                       class="w-[20px] h-[20px] my-0.5 bg-white rounded-full border border-gray-300"
                     />
-                    <div class="grow">
+                    <div class="flex-1">
                       <div class="flex justify-between items-center gap-3 text-green-500">
                         <span
                           class="font-semibold"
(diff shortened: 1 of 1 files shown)


pages/store/[id].vue
```diff
@@ -157,7 +157,7 @@
                       v-else
                       class="w-[20px] h-[20px] my-0.5 bg-white rounded-full border border-gray-300"
                     />
-                    <div class="grow">
+                    <div class="flex-1">
```
Comment (nwingt) [inline pages/store/[id].vue]: I think we can keep `grow`

Comment (AuroraHuang22) [reply pages/store/[id].vue]: 截圖 2025-06-07 13 50 34

Keeping `grow` makes the unselected checkbox shrink.
Or do you want to keep `grow` but fix the width of the checkbox?

pages/store/[id].vue
```diff
@@ -157,7 +157,7 @@
                       v-else
                       class="w-[20px] h-[20px] my-0.5 bg-white rounded-full border border-gray-300"
```
Comment (nwingt) [inline pages/store/[id].vue]:
```suggestion
                      class="shrink-0 w-[20px] h-[20px] my-0.5 bg-white rounded-full border border-gray-300"
```

pages/store/[id].vue
Comment (nwingt) [reply pages/store/[id].vue]: should add `shrink-0` to the unselected checkbox

```suggestion
                    <div class="grow">
```

nwingt (changes requested):

nwingt (approved):

PR #49, opened 2025-06-11
✨ Add account registration

nwingt:
Depends on https://github.com/likecoin/likecoin-api-public/pull/957 && https://github.com/likecoin/likecoin-api-public/pull/958


stores/account.ts
```diff
@@ -62,6 +63,115 @@ export const useAccountStore = defineStore('account', () => {
...
+    loginMethod,
+    magicUserId,
+  }: {
+    walletAddress: string
+    email?: string
+    loginMethod: string
+    magicUserId?: string
+  }) {
+    // Loop until registration is successful or user cancels
+    const tempAccountId = await generateAccountIdFromWalletAddress(walletAddress)
```
williamchong: Should we check if this accountId is used first? Can we regenerate a different `AccountId` if hash collision somehow occur?

i18n/locales/en.json
```diff
@@ -8,14 +9,15 @@
...
-  "account_page_liker_id": "Liker ID",
   "account_page_locale": "Language",
   "account_page_login": "Login",
   "account_page_logout": "Logout",
   "account_page_publish_book": "Publish Book",
   "account_page_settings_and_help_title": "Settings",
   "account_page_title": "Account",
   "account_page_update_settings_error": "Unable to update account settings.",
+  "account_register_error_email_already_used": "[{email}] was bound to [{walletAddress}] and cannot be used to register a new account. Please login with your wallet instead.",
+  "account_register_error_invalid_account_id": "Account ID {id} is not accepted.",
```
```suggestion
  "account_register_error_invalid_account_id": "Account ID {id} is invalid.",
```

stores/account.ts
```diff
@@ -84,35 +194,55 @@ export const useAccountStore = defineStore('account', () => {
...
+      const loginMethod = connector.id
+      if (loginMethod === 'magic' && 'magic' in connector) {
         const magic = connector.magic as Magic
         try {
           const userInfo = await magic.user.getInfo()
           if (userInfo.email) {
             email = userInfo.email
           }
+          if (userInfo.issuer) {
+            magicUserId = userInfo.issuer
```
williamchong, line 271: Better send idToken and verify the whole idToken in server

server/api/register.post.ts
```diff
@@ -0,0 +1,127 @@
...
+      message: 'REGISTER_INVALID_BODY',
+    })
+  }
+
+  const config = useRuntimeConfig()
+  try {
+    await $fetch(`${config.public.likeCoinAPIEndpoint}/users/new`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
```
williamchong: Not sure if this is really required

components/RegistrationModal.vue
```diff
@@ -0,0 +1,123 @@
...
+    :title="$t('registration_modal_title')"
+    :ui="{
+      content: 'max-w-[348px]',
+      body: 'space-y-4',
+      footer: 'flex justify-end',
+    }"
+  >
+    <template #body>
+      <UFormField
+        v-if="!props.isAccountIdHidden"
```
williamchong: Not sure if we want to show the field if we hit any `accountIdError`

server/api/register.post.ts
```diff
@@ -0,0 +1,126 @@
...
+      displayName: string
+      description: string
+      avatar: string
+    }>(`${config.public.likeCoinAPIEndpoint}/users/addr/${body.walletAddress}/min`, {
+      query: { ts: Date.now() },
+    })
+    likerId = userInfoRes.user
+    displayName = userInfoRes.displayName
+    avatar = userInfoRes.avatar
+    description = userInfoRes.description
```
williamchong: Please add `isLikerPlus`, or may as well refactor this fetch/refresh liker id info function if you have time

stores/account.ts
```diff
@@ -84,35 +243,57 @@ export const useAccountStore = defineStore('account', () => {
...
 
+      // Check if the wallet address or email is already registered
+      let isRegistered = await checkIsRegistered({ walletAddress, email, magicDIDToken })
+      if (!isRegistered) {
+        // If not registered, proceed to registration
+        isRegistered = await register({ walletAddress, email, loginMethod, magicUserId, magicDIDToken })
+        if (!isRegistered) {
+          // User canceled the registration
+          return
+        }
```
williamchong, line 288: Do we still need to sign for login if user just `register()`ed?

stores/account.ts
```diff
@@ -62,6 +63,164 @@ export const useAccountStore = defineStore('account', () => {
...
+          ? {
+              accountId: tempAccountId,
+              email,
+            }
+          : await registrationFormModal.open({
+            accountId: tempAccountId,
+            isAccountIdHidden: true,
+            email: payload?.email || '',
+            isDisplayNameHidden: true,
+          })
```
williamchong: Seems no way to exit the loop if `email` is provided and the error is one of the handled cases/
Can we just loop for `registrationFormModal.open` case, or don't loop in store and loop from caller (page/component) instead?

components/RegistrationModal.vue
```diff
@@ -0,0 +1,131 @@
...
+
+  emit('close', {
+    accountId: accountId.value,
+    email: email.value,
+    displayName: displayName.value,
+  })
+}
+
+function handleClose() {
+  useTrackEvent('register_liker_id_close_button_click')
```
williamchong:
```suggestion
  useLogEvent('register_liker_id_close_button_click')
```

components/RegistrationModal.vue
```diff
@@ -0,0 +1,131 @@
...
+
+watch(accountIdError, (newValue) => {
+  // If accountId has an error, show the accountId field if it was hidden
+  if (newValue && isAccountIdHidden.value) {
+    isAccountIdHidden.value = false
+  }
+})
+
+function handleRegisterButtonClick() {
+  useTrackEvent('register_button_click')
```
williamchong:
```suggestion
  useLogEvent('register_button_click')
```

williamchong (changes requested):
nice hash function

stores/account.ts
nwingt: yes, below will call `authorize` action

stores/account.ts
nwingt: How about this https://github.com/likecoin/3ook-com/pull/49/commits/ad147f94aa25026f1d4d62e5eee4ab9f95e135e9

williamchong (approved):

PR #63, opened 2025-06-13
[TAS-3899] 📈 Add meta pixel

williamchong:


composables/use-log-event.ts
```diff
@@ -2,17 +2,55 @@ interface EventParams {
...
+        order_id: paymentId,
+        content_type: items ? 'product' : undefined,
+        contents: Array.isArray(items)
+          ? items.map(i => ({
+              id: i.id,
+              quantity: i.quantity || 1,
+            }))
+          : undefined,
+        content_ids: Array.isArray(items) ? items.map(i => i.id) : undefined,
+        eventID,
```
nwingt: why mixing snake case and camelCase?

williamchong: https://developers.facebook.com/docs/meta-pixel/implementation/conversion-tracking?locale=zh_HK?
https://developers.facebook.com/docs/marketing-api/conversions-api/deduplicate-pixel-and-server-events/
inside and outside of payload i guess..

williamchong: Moved `eventID` to 3rd param

composables/use-log-event.ts
```diff
@@ -2,17 +2,54 @@ interface EventParams {
...
+      search: 'Search',
+    }
+    if (eventNameMapping[eventName]) {
+      const {
+        transaction_id: paymentId,
+        value,
+        currency,
+        items,
+      } = eventParams
+      const eventID = paymentId ? `${eventName}_${paymentId}` : undefined
```
nwingt: Can we keep `Id` consistent?
```suggestion
      const eventId = paymentId ? `${eventName}_${paymentId}` : undefined
```

williamchong: Updated, we still need to do `eventID: eventId` in the actual payload though

nwingt (approved):

PR #68, opened 2025-06-13
📈 Add user id properties in logger

williamchong:


composables/use-log-event.ts
```diff
@@ -1,8 +1,11 @@
+import { setUser } from '@sentry/nuxt'
```
nwingt: Prevent possible name collision 
```suggestion
import { setUser as setSentryUser } from '@sentry/nuxt'
```

composables/use-log-event.ts
```diff
@@ -23,3 +26,71 @@ export default function useLogEvent(eventName: string, eventParams?: EventParams
     }
   }
 }
+
+export function useLogSetUser(user: {
```
nwingt: Why not move this composable to a separate file i.e. `use-set-log-user.ts`?
```suggestion
export function useSetLogUser(user: {
```

composables/use-log-event.ts
```diff
@@ -23,3 +26,71 @@ export default function useLogEvent(eventName: string, eventParams?: EventParams
...
+  email?: string
+  evmWallet?: string
+  likeWallet?: string
+  loginMethod?: string
+  displayName?: string
+  avatar?: string
+  id?: string
+} | null) {
```
nwingt: Is the type equal to: https://github.com/likecoin/3ook-com/blob/c7897fda2b44fcdb20e18a7f6d018d62fd86f3b5/types/nuxt-auth-utils.d.ts#L2-L15

composables/use-log-event.ts
williamchong: these functions have similar structure
```
if... setSentry...
if... setGa...
if... setMeta...
if... setCrisp...
```
seems placing in same file is easier to maintain
renaming file to `use-logger` to avoid confusion

nwingt (approved):

PR #70, opened 2025-06-14
[TAS-3654] ✨ Add migration notice

nwingt:
### Before
Screenshot 2025-06-15 at 01 33 03

### After
Screenshot 2025-06-15 at 01 30 38


i18n/locales/zh-Hant.json
```diff
@@ -66,6 +66,9 @@
   "footer_terms": "使用條款",
   "login_button": "登入",
   "login_panel_continue_with_email": "以電郵登入",
+  "login_panel_migration_notice": "原 Liker Land 用戶可{here}升級原有帳戶及資料",
+  "login_panel_migration_notice_here": "於此",
+  "login_panel_migration_notice_link": "https://link.liker.land/v3-migration-zh",
```
williamchong: Go straight to migration ui until we have actual doc?

nwingt: it is a redirect-able short URL

williamchong (approved):

PR #71, opened 2025-06-15
[TAS-3956] 🐛 Fix author description

williamchong:


nwingt (approved):

composables/use-book-info.ts
```diff
@@ -29,7 +29,9 @@ export default function ({ nftClassId = '' }: { nftClassId?: string } = {}) {
   })
 
   const authorDescription = computed(() => {
-    return bookstoreInfo.value?.author?.description || ''
+    const author = bookInfo.author.value
+    if (typeof author === 'object' && 'description' in author) return author.description
+    return ''
```
nwingt, line 34: As fallback
```suggestion
    return bookstoreInfo.value?.author?.description || ''
```

nwingt (changes requested):

composables/use-book-info.ts
williamchong: if `bookstoreInfo` has a diff value most likely it is just outsync now

PR #76, opened 2025-06-16
[TAS-3959] 🥅 Logout user when permissions are insufficient

nwingt:


composables/use-error-handler.ts
```diff
@@ -34,12 +34,13 @@ export default function () {
     // Generic error handling
     if (!handler) {
       switch (rawErrorMessage) {
+        case 'INSUFFICIENT_PERMISSION':
         case 'TOKEN_EXPIRED':
           handler = {
             description: $t('error_token_expired'),
             onClose: () => {
               accountStore.logout()
```
williamchong: Do you need to await for this before `navigateTo`?

nwingt: Update with UI improvements

i18n/locales/en.json
```diff
@@ -1,5 +1,9 @@
 {
   "account_export_private_key_not_supported": "Export private key is not supported.",
+  "account_logged_in": "Login successfully",
```
```suggestion
  "account_logged_in": "Login successful",
```

composables/use-error-handler.ts
```diff
@@ -34,12 +34,13 @@ export default function () {
...
       switch (rawErrorMessage) {
+        case 'INSUFFICIENT_PERMISSION':
         case 'TOKEN_EXPIRED':
           handler = {
             description: $t('error_token_expired'),
-            onClose: () => {
-              accountStore.logout()
-              navigateTo(localeRoute({ name: 'account' }))
+            onClose: async () => {
+              await accountStore.logout()
```
williamchong: Do we need to catch here? Like if `logout()` throws error because we aren't actually logged in or sth

williamchong (changes requested):

composables/use-error-handler.ts
nwingt: Updated

williamchong (approved):

PR #78, opened 2025-06-16
👽 Update FAQ & publish links

nwingt:


williamchong (approved):
copilot says `?utm_source=3ookcom&utm_medium=referral&utm_campaign=account_page&utm_content=faq_button` but i am also good with this until we find need to change

PR #99, opened 2025-06-18
[TAS-3998] 🐛 Fix bookshelf infinite loading stop prematurely

williamchong:


stores/bookshelf.ts
```diff
@@ -31,11 +31,13 @@ export const useBookshelfStore = defineStore('bookshelf', () => {
       isFetching.value = true
       let res: FetchLikeCoinChainNFTsResponseData | FetchLegacyLikeCoinChainNFTClassesResponseData
       const key = isRefresh ? undefined : nextKey.value?.toString()
+      const limit = 100
```
nwingt: Can move to function params?

pages/shelf.vue
```diff
@@ -129,9 +129,12 @@ watch(
...
   () => shouldLoadMore.value,
-  (shouldLoadMore) => {
-    if (shouldLoadMore) {
-      bookshelfStore.fetchItems()
+  async (loadMore) => {
+    if (loadMore) {
+      do {
+        await bookshelfStore.fetchItems()
+        await sleep(100)
+      } while (bookshelfStore.nextKey && shouldLoadMore.value)
```
nwingt: when load more is triggered, it keeps fetching till the end

williamchong: The scroll bar is stuck at bottom causing visibility to be always true, added a scrollBy hack

nwingt (approved):

PR #100, opened 2025-06-18
[TAS-3987] ✨ Fetch NFT metadata in when fetching bookshelf items

williamchong:


stores/nft.ts
```diff
@@ -110,6 +111,9 @@ export const useNFTStore = defineStore('nft', () => {
 
   async function lazyFetchNFTClassAggregatedMetadataById(nftClassId: string, { exclude = [] }: FetchLikeCoinNFTClassAggregatedMetadataOptions = {}) {
     const excludedOptions: LikeCoinNFTClassAggregatedMetadataOptionKey[] = exclude
+    if (nftClassChainMetadataByIdMap.value[nftClassId]) {
+      excludedOptions.push('class_chain')
+    }
```
nwingt: need to update the non-lazy one?

stores/bookshelf.ts
```diff
@@ -49,6 +51,10 @@ export const useBookshelfStore = defineStore('bookshelf', () => {
               name: item.name,
               owner_address: item.owner_address,
             })
+
+            nftStore.lazyFetchNFTClassChainMetadataById(nftClassId).then((metadata) => {
+              nftStore.addNFTClassMetadata(nftClassId, metadata)
```
nwingt: Should this be called within `fetchNFTClassChainMetadataById()`?

stores/nft.ts
williamchong: Rewrote to reuse existing fields

nwingt (approved):

PR #101, opened 2025-06-19
🚸 Fix book cover placeholder is not clickable

williamchong:
not sure if I can always use `h-full` even for normal case

### DIFF
--- a/components/BookCover.vue
+++ b/components/BookCover.vue
@@ -14,6 +14,7 @@
         'w-full h-full': hasError,
       }"
       :to="props.to"
+      @click="emit('click', $event)"
     >
       <div
         v-if="hasShadow"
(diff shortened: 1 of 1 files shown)


AuroraHuang22 (approved):

components/BookCover.vue
```diff
@@ -135,6 +135,9 @@ const coverClass = computed(() => {
   else {
     classes.push('pointer-events-none')
   }
+  if (hasError.value) {
+    classes.push('h-full')
```
nwingt: but this has `absolute` and `inset-0`, how does it help?

williamchong: image

without h-full

image

with h-full

nwingt:
@williamchong https://github.com/williamchong/liker-land-v3/pull/1

The `` is not meant to be clickable, its parent (the `<a/>`) is


https://github.com/user-attachments/assets/8155c125-69a3-4599-bf3f-efd97889bff3

PR #110, opened 2025-06-23
📈 Add plus purchase events

williamchong:


pages/plus/success.vue
```diff
@@ -52,6 +59,26 @@ onMounted(async () => {
...
+          quantity: 1,
+        }],
+      })
+      await navigateTo(localeRoute({
+        name: getRouteBaseName(route),
+        query: {
+          ...route.query,
+          redirect: undefined,
+        },
+      }), { replace: true })
```
nwingt, line 80: Should return here?
```suggestion
      }), { replace: true })
      return
```

williamchong: This is just to get rid of `redirct=1` qs to prevent duplicated event logging, other logic is suppose to continue

nwingt (approved):

components/PaywallModal.vue
```diff
@@ -12,7 +13,6 @@
         class="absolute z-10 top-4 right-4 cursor-pointer"
         variant="link"
         size="md"
-        @click="handleCloseButtonClick"
```
nwingt, line 15: Close button not working

PR #112, opened 2025-06-23
📈 Add ga session_id fetch and recover logic

williamchong:


composables/use-ga-session-ids.ts
```diff
@@ -0,0 +1,23 @@
+export function useGaSessionIds() {
```
nwingt:
```suggestion
export function useGoogleAnalytics() {
```

pages/store/[id].vue
```diff
@@ -347,6 +347,7 @@ const { loggedIn: hasLoggedIn, user } = useUserSession()
 const accountStore = useAccountStore()
 const nftStore = useNFTStore()
 const { handleError } = useErrorHandler()
+const { gaClientId, gaSessionId } = useGaSessionIds()
```
Comment (nwingt) [inline pages/store/[id].vue]:
```suggestion
const { gaClientId, gaSessionId } = useGoogleAnalytics()
```

plugins/gtag.ts
```diff
@@ -0,0 +1,12 @@
+export default defineNuxtPlugin(() => {
+  const { gtag } = useGtag()
+  const googleAnalyticsTrackingId = useRuntimeConfig().public.googleAnalyticsTrackingId as string
```
nwingt, line 3: image

```suggestion
  const googleAnalyticsTrackingId = useRuntimeConfig().public.googleAnalyticsTrackingId
```

nwingt (changes requested):

nwingt (approved):

PR #113, opened 2025-06-23
📈 Add gad conversion id support

williamchong:


composables/use-logger.ts
```diff
@@ -9,6 +9,14 @@ interface EventParams {
 export function useLogEvent(eventName: string, eventParams: EventParams = {}) {
   try {
     useTrackEvent(eventName, eventParams)
+    const AD_CONVERSION_ID = useRuntimeConfig().public.AD_CONVERSION_ID as string
```
nwingt: Seems you havent inject runtime config in `nuxt.config.ts`

nwingt (changes requested):

nwingt (approved):

PR #121, opened 2025-06-25
[TAS-4028] 🐛 Fix incorrect fileIndex used for download

AuroraHuang22:


composables/use-book-download.ts
```diff
@@ -42,7 +42,7 @@ export function useBookDownload() {
   }: {
     nftClassId: string
     nftId: string
-    fileIndex?: string
+    fileIndex?: number
```
nwingt, line 45: please update the default value as well

nwingt (changes requested):

composables/use-book-download.ts
AuroraHuang22: [🐛 Fix incorrect fileIndex used for download](https://github.com/likecoin/3ook-com/pull/121/commits/ccc44040e9309f4ffff8b20ac5ebf1b8759a623b)

nwingt (approved):

PR #123, opened 2025-06-25
[TAS-3996] ✨ Add token-id page as redirect

williamchong:


composables/use-book-info.ts
```diff
@@ -168,8 +168,8 @@ export default function ({ nftClassId = '' }: { nftClassId?: string } = {}) {
   const firstUserOwnedNFTId = computed(() => userOwnedNFTIds.value[0])
 
   const productPageRoute = computed(() => localeRoute({
-    name: 'store-id',
-    params: { id: nftClassId },
+    name: 'store-classId',
```
nwingt: Prefer fullname 
```suggestion
    name: 'store-nftClassId',
```

pages/store/[classId]/[tokenId].vue
Comment (nwingt) [inline pages/store/[classId]/[tokenId].vue:1]: On second thought, I think we should use `nftId` instead of `tokenId`
since we use `nftId` throughout the codebase

nwingt (approved):

PR #128, opened 2025-06-27
✨ Introduce tags

nwingt:


ckxpress:
Screenshot 2025-06-30 at 16 38 33

is this layout temporary? cos i remember it should be a list of tags going out horizontally in a row?

nwingt:
> is this layout temporary? cos i remember it should be a list of tags going out horizontally in a row?

@ckxpress Yes, I will deal with this later because the tag list UI is troublesome on mobile devices

pages/store/index.vue
```diff
@@ -54,21 +103,108 @@
...
+  get: () => getRouteQuery('tag', TAG_LISTING),
+  set: async (id) => {
+    await navigateTo(localeRoute({
+      name: 'store',
+      query: {
+        ...route.query,
+        // NOTE: Remove the tag query if it is the listing tag
+        tag: id === TAG_LISTING ? undefined : id,
+      },
+    }))
```
williamchong, line 130: Not sure if should `replace: true`

williamchong (approved):

pages/store/index.vue
nwingt: I think each tag should be treated as a page

PR #130, opened 2025-06-27
[TAS-4259] ✨ Implement TTS in PDF viewer

williamchong:

PR #136, opened 2025-06-30
♻️ Refactor tts into composable with double buffer

williamchong:


composables/use-text-to-speech.ts
```diff
@@ -0,0 +1,181 @@
...
+  onEnd?: (element: TextContentElement) => void
+  onError?: (error: Event) => void
+  onPageChange?: () => void
+  checkIfNeedPageChange?: (element: TextContentElement) => boolean
+}
+
+export function useTextToSpeech(options: TTSOptions = {}) {
+  const { user } = useUserSession()
+  const subscription = useSubscription()
+  const { nftClassId } = useReader()
```
nwingt: https://github.com/likecoin/3ook-com/pull/130#discussion_r2172354202

composables/use-text-to-speech.ts
```diff
@@ -0,0 +1,181 @@
...
+  id: string
+  text: string
+}
+
+interface TTSOptions {
+  onPlay?: (element: TextContentElement) => void
+  onEnd?: (element: TextContentElement) => void
+  onError?: (error: Event) => void
+  onPageChange?: () => void
+  checkIfNeedPageChange?: (element: TextContentElement) => boolean
```
nwingt: Screenshot 2025-07-01 at 09 54 45

nwingt (approved):

PR #137, opened 2025-06-30
✨ Add TTS voice selection support via temp UI

williamchong:


pages/reader/epub.vue
```diff
@@ -314,14 +314,18 @@ onMounted(async () => {
...
+  { label: '粵0', value: { language: 'zh-HK', voiceId: '0' } },
+  { label: '粵1', value: { language: 'zh-HK', voiceId: '1' } },
+  { label: '國0', value: { language: 'zh-TW', voiceId: '0' } },
+  { label: '國1', value: { language: 'zh-TW', voiceId: '1' } },
+  { label: 'En0', value: { language: 'en-US', voiceId: '0' } },
+  { label: 'En1', value: { language: 'en-US', voiceId: '1' } },
 ]
-const ttsLanguage = ref('zh-HK')
-watch(ttsLanguage, (newLanguage, oldLanguage) => {
+const ttsLanguageVoice = ref({ language: 'zh-HK', voiceId: '0' })
```
nwingt: Can `<USelect/>` support object value?

williamchong: image
image


Seems to work since it is not actually `<select>`, not sure if desired or undefined behaviour

williamchong: In type this falls into `AcceptableValue`
image
image

nwingt (approved):

PR #139, opened 2025-07-01
🐛 Fix NFT type filtering in items computation

AuroraHuang22:


stores/bookshelf.ts
```diff
@@ -11,9 +11,14 @@ export const useBookshelfStore = defineStore('bookshelf', () => {
     return Object.values(nftByNFTClassIds.value[nftClassId] || {})
   })
 
+  type Metadata = { '@type'?: string, 'type'?: string }
```
nwingt, line 14: We have a type definition, we can add type to it with a comment that this is legacy
https://github.com/likecoin/3ook-com/blob/9a377454b0bfb8b20723ed3bee0b80cc0ea1b02f/types/nft.d.ts#L1-L9

nwingt (changes requested):

williamchong:
I suggest not to merge this PR if we can confirm production data is free of `type`

williamchong:
Closing this since all `type` should be fixed in `prod` now

PR #142, opened 2025-07-02
✨ Allow viewing someone else's bookshelf

nwingt:


pages/shelf/[[walletAddress]].vue
```diff
@@ -2,9 +2,35 @@
   <div>
     <AppHeader />
 
+    <header
+      v-if="isSomeoneElseBookshelf"
```
Comment (williamchong) [inline pages/shelf/[[walletAddress]].vue]:
```suggestion
      v-if="!isOwnedBookshelf"
```
I think flipped condition would be cleaner in naming, but just suggestion

pages/shelf/[[walletAddress]].vue
```diff
@@ -72,49 +98,96 @@
...
-useHead({
-  title: $t('shelf_page_title'),
+useHead(() => {
+  const meta = []
+  if (isSomeoneElseBookshelf.value) {
+    meta.push({
+      name: 'robots',
+      content: 'noindex',
+    })
+  }
```
Comment (williamchong) [inline pages/shelf/[[walletAddress]].vue]: We can safely set `noIndex` for all cases, since `/shelf` requires login if owned

pages/shelf/[[walletAddress]].vue
```diff
@@ -131,6 +204,7 @@ function handleBookshelfItemOpen({
   useLogEvent('shelf_open_book', {
     content_type: type,
     nft_class_id: nftClassId,
+    evm_wallet: walletAddress.value,
```
Comment (williamchong) [inline pages/shelf/[[walletAddress]].vue]: this is basically user_id, not sure if we want to log this

pages/shelf/[[walletAddress]].vue
```diff
@@ -144,6 +218,7 @@ function handleBookshelfItemDownload({
   useLogEvent('shelf_download_book', {
     content_type: type,
     nft_class_id: nftClassId,
+    evm_wallet: walletAddress.value,
```
Comment (williamchong) [inline pages/shelf/[[walletAddress]].vue]: this is basically user_id, not sure if we want to log this

williamchong (approved):

PR #143, opened 2025-07-02
✨ Add TTS playback speed and skip support

williamchong:


composables/use-text-to-speech.ts
```diff
@@ -25,15 +25,27 @@ export function useTextToSpeech(options: TTSOptions = {}) {
...
+  const currentAudioTimeout = ref<NodeJS.Timeout | null>(null)
   const currentBufferIndex = ref<0 | 1>(0)
   const currentElementIndex = ref(0)
   const textContentElements = ref<TextContentElement[]>([])
+  const currentTextElement = computed(() => {
+    return textContentElements.value[currentElementIndex.value]
+  })
+  const currentTextContent = computed(() => {
+    return currentTextElement.value?.text || ''
+  })
```
nwingt: The naming is a bit confusing. I propose the following changes (with the help of Copilot):
```
currentElementIndex → currentTTSSegmentIndex
textContentElements → ttsSegments
currentTextElement → currentTTSSegment
currentTextContent → currentTTSSegmentText

type TextContentElement  → type TTSSegment
```

- **currentTTSSegmentIndex:**  
  The index of the current TTS segment being processed or spoken.

- **ttsSegments:**  
  The array containing all TTS segments (split by punctuations).

- **currentTTSSegment:**  
  The currently active TTS segment object, derived from `ttsSegments` using the index.

- **currentTTSSegmentText:**  
  The text string of the current TTS segment.

composables/use-text-to-speech.ts
```diff
@@ -180,6 +202,35 @@ export function useTextToSpeech(options: TTSOptions = {}) {
...
+    if (!isTextToSpeechOn.value) return
+    const activeAudio = audioBuffers.value[currentBufferIndex.value]
+    activeAudio?.pause()
+    if (currentTextElement.value) options.onEnd?.(currentTextElement.value)
+    playNextElement()
+  }
+
+  function skipBackward() {
+    if (!isTextToSpeechOn.value) return
+    const activeAudio = audioBuffers.value[currentBufferIndex.value]
```
nwingt: Should we add logs here?

nwingt (approved):

PR #144, opened 2025-07-02
[TAS-3989] 💩 Fetch tags in client side

nwingt:


pages/store/index.vue
```diff
@@ -236,13 +255,30 @@ async function fetchItems({ isRefresh = false } = {}) {
...
 onMounted(async () => {
+  await fetchTags()
+  if (!tag.value) {
+    throw createError({
+      statusCode: 404,
+      message: $t('error_page_not_found'),
+      fatal: true,
+    })
+  }
   await fetchItems({ isRefresh: true })
```
williamchong: Can we do this in `Promise.all` or do they have dependency?

nwingt: Updated

williamchong (approved):

PR #149, opened 2025-07-03
[TAS-4063] 🚸 Add sold out indication for pricing items

AuroraHuang22:
before
image
after
image

### DIFF
--- a/pages/store/[nftClassId]/index.vue
+++ b/pages/store/[nftClassId]/index.vue
@@ -166,7 +166,12 @@
                           class="font-semibold"
                           v-text="item.name"
                         />
-                        <span>
+                        <span
+                          v-if="item.isSoldOut"
+                          class="text-sm font-semibold"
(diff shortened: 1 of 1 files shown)


pages/store/[nftClassId]/index.vue
```diff
@@ -284,7 +289,15 @@
           'z-10',
         ]"
       >
-        <span class="text-green-500">
+        <span
+          v-if="selectedPricingItem?.isSoldOut"
+          class="text-gray-400"
+          v-text="selectedPricingItem.name"
+        />
```
Comment (nwingt) [inline pages/store/[nftClassId]/index.vue]: I am concerned there is not enough space for the name in mobile

Maybe keep showing the price but in gray color?

Comment (AuroraHuang22) [reply pages/store/[nftClassId]/index.vue]: 👌 Keep the existing UI for mobile

nwingt (approved):

PR #151, opened 2025-07-03
⚡️ Preload book cover image for store page

williamchong:


nwingt (commented):
Need to consider retina screen (2x)

williamchong:
Just the preload for LCP then

nwingt (approved):

PR #153, opened 2025-07-04
👽️ Set stream_options in TTS API

williamchong:


nwingt (approved):
What does this do?

williamchong:
> * exclude_aggregated_audio

image

Chinese doc has this extra option that skip the summary block at the end, which is not useful in our streamed case

PR #156, opened 2025-07-04
⚡️ Preload store API in store index

williamchong:


stores/bookstore.ts
```diff
@@ -1,3 +1,5 @@
+import { getTimestampRoundedToMinute } from '../utils'
```
nwingt: not auto import?

williamchong: `getTimestampRoundedToMinute` is not found error randomly shows up, not sure if related to dev vs prod or ssr vs client

PR #158, opened 2025-07-04
[TAS-4085] ⚡️ Lazy load image in paginated list

williamchong:


composables/use-paginated-grid.ts
```diff
@@ -69,5 +69,6 @@ export default function usePaginatedGrid(props: {
   return {
     gridClasses,
     getGridItemClassesByIndex,
+    COLUMN_MAX,
```
nwingt: we have a computed prop `columnMax`

nwingt (changes requested):

nwingt (approved):

PR #169, opened 2025-07-09
[TAS-4123]💄 Use slideover on mobile to prevent toolbar overflow

AuroraHuang22:
截圖 2025-07-09 15 38 09
截圖 2025-07-09 15 38 24
截圖 2025-07-09 15 38 17

### DIFF
--- a/pages/reader/epub.vue
+++ b/pages/reader/epub.vue
@@ -59,79 +59,111 @@
             />
             <template v-if="!bookInfo.isAudioHidden.value">
               <template v-if="!isTextToSpeechOn || !isTextToSpeechPlaying">
-                <UButton
-                  class="laptop:hidden"
-                  icon="i-material-symbols-headphones-rounded"
-                  variant="ghost"
-                  :disabled="isReaderLoading"
-                  @click="startTextToSpeech"
-                />
-                <UButton
-                  class="max-laptop:hidden"
-                  icon="i-material-symbols-headphones-rounded"
-                  :label="$t('reader_text_to_speech_button')"
-                  variant="ghost"
-                  :disabled="isReaderLoading"
-                  @click="startTextToSpeech"
-                />
+                <UButtonGroup>
+                  <UButton
+                    class="laptop:hidden"
+                    icon="i-material-symbols-headphones-rounded"
+                    variant="outline"
+                    color="neutral"
+                    :disabled="isReaderLoading"
+                    @click="startTextToSpeech"
+                  />
+                  <UButton
+                    class="max-laptop:hidden"
+                    icon="i-material-symbols-headphones-rounded"
+                    :label="$t('reader_text_to_speech_button')"
+                    variant="outline"
+                    color="neutral"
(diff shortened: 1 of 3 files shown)


pages/reader/epub.vue
```diff
@@ -91,27 +91,46 @@
...
+                  :close="{
+                    color: 'neutral',
+                    variant: 'outline',
+                    class: 'rounded-full',
+                  }"
+                  side="bottom"
+                >
+                  <UButton
+                    v-if="isShowTextToSpeechOptions"
+                    icon="i-lucide-settings-2"
```
nwingt: please use icon from material-symbols

pages/reader/epub.vue
```diff
@@ -91,27 +91,46 @@
...
+                    variant="ghost"
+                  />
+                  <template #body>
+                    <div class="flex items-center gap-2">
```
nwingt: Can remove this `<div/>`, and add class to `<USlideover>` ui props

pages/reader/epub.vue
```diff
@@ -91,27 +91,46 @@
...
-                  @click="skipForward"
-                />
+                <USlideover
+                  :title="$t('reader_tts_options_button')"
+                  :close="{
+                    color: 'neutral',
+                    variant: 'outline',
+                    class: 'rounded-full',
+                  }"
+                  side="bottom"
```
nwingt:
```suggestion
                  side="bottom"
                  :overlay="false"
```

nwingt (changes requested):

pages/reader/epub.vue
williamchong: Suggest to use `side="top"` with `overlay="false"`

pages/reader/epub.vue
```diff
@@ -91,27 +91,46 @@
...
+                >
+                  <UButton
+                    v-if="isShowTextToSpeechOptions"
+                    icon="i-lucide-settings-2"
+                    variant="ghost"
+                  />
+                  <template #body>
```
williamchong: Add Play/Pause button in this toolbar also?

pages/reader/epub.vue
```diff
@@ -91,27 +91,46 @@
...
-                  :items="ttsPlaybackRateOptions"
-                />
-                <UButton
-                  icon="i-material-symbols-skip-next-rounded"
-                  variant="ghost"
-                  :disabled="!isTextToSpeechOn"
-                  @click="skipForward"
-                />
+                <USlideover
+                  :title="$t('reader_tts_options_button')"
```
williamchong: Would be nice if we can make the title go away, such that the slide over would just cover the height of original header, then user can still read the book content below when using TTS control
Not sure if possible in `<USlideover>` though
image

williamchong (changes requested):
see if @nwingt think my comment is sensible

nwingt (approved):

williamchong (approved):

PR #170, opened 2025-07-09
[TAS-4066] 💩 Mitigate rtl EPUB issues

nwingt:


williamchong:
Mitigate with https://github.com/likecoin/3ook-com/pull/176 instead

PR #178, opened 2025-07-09
[TAS-4092] 🚸 Use contract function for claim flow

nwingt:


pages/store/claim.vue
```diff
@@ -133,6 +133,7 @@ async function checkItemsDelivery() {
   // TODO: Handle multiple items
   try {
     await Promise.all([
+      bookshelfStore.fetchNFTByNFTClassIdAndOwnerWalletAddress(nftClassId.value as string, user.value?.evmWallet as string),
```
williamchong: I think we need to set a max time to wait for actual indexer response, before we just use the result on chain
We need indexer due to transaction memo used in book signature page

We also need to modify book url in reader to no enable `custom message` if we are not using result from indexer

nwingt: Updated

williamchong (approved):

PR #181, opened 2025-07-10
🚸 Support mediaSession in TTS

williamchong:
image

Set name/cover/control in widget

### DIFF
--- a/composables/use-text-to-speech.ts
+++ b/composables/use-text-to-speech.ts
@@ -59,39 +59,44 @@ export function useTextToSpeech(options: TTSOptions = {}) {
   })
 
   function setupMediaSession() {
-    if ('mediaSession' in navigator) {
-      navigator.mediaSession.metadata = new MediaMetadata({
-        title: toValue(bookName),
-        album: toValue(bookChapterName) || toValue(bookName),
-        artist: toValue(bookAuthorName),
-        artwork: bookCoverSrc
-          ? [
-              {
-                src: toValue(bookCoverSrc),
-              },
-            ]
-          : undefined,
-      })
+    try {
+      if ('mediaSession' in navigator) {
+        navigator.mediaSession.metadata = new MediaMetadata({
+          title: toValue(bookName),
+          album: toValue(bookChapterName) || toValue(bookName),
+          artist: toValue(bookAuthorName),
+          artwork: bookCoverSrc
+            ? [
+                {
+                  src: toValue(bookCoverSrc),
+                },
+              ]
+            : undefined,
+        })
 
-      navigator.mediaSession.setActionHandler('play', () => {
-        startTextToSpeech()
-      })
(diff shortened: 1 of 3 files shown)


composables/use-text-to-speech.ts
```diff
@@ -10,10 +10,20 @@ interface TTSOptions {
   onError?: (error: Event) => void
   onPageChange?: (direction?: number) => void
   checkIfNeededPageChange?: (element: TTSSegment) => boolean
+  bookInfo?: {
+    name?: string | Ref<string> | ComputedRef<string>
+    author?: string | Ref<string> | ComputedRef<string>
+    image?: string | Ref<string> | ComputedRef<string>
+  }
```
nwingt: prefer not nested 

bookName
bookAuthorName
bookCoverSrc

composables/use-text-to-speech.ts
```diff
@@ -48,6 +58,49 @@ export function useTextToSpeech(options: TTSOptions = {}) {
     return currentTTSSegment.value?.text || ''
   })
 
+  function setupMediaSession() {
+    if ('mediaSession' in navigator) {
+      navigator.mediaSession.metadata = new MediaMetadata({
+        title: toValue(bookName),
+        album: toValue(bookChapterName) || toValue(bookName),
+        artist: toValue(bookAuthorName),
+        artwork: bookCoverSrc
```
```suggestion
        artwork: toValue(bookCoverSrc)
```

pages/reader/epub.vue
```diff
@@ -433,6 +433,10 @@ const {
   stopTextToSpeech,
 } = useTextToSpeech({
   nftClassId: nftClassId.value,
+  bookName: bookInfo.name,
+  bookChapterName: computed(() => activeNavItemLabel.value),
```
nwingt: Isn't `activeNavItemLabel` a computed prop?

composables/use-text-to-speech.ts
```diff
@@ -48,6 +58,54 @@ export function useTextToSpeech(options: TTSOptions = {}) {
...
   })
 
+  function setupMediaSession() {
+    try {
+      if ('mediaSession' in navigator) {
+        navigator.mediaSession.metadata = new MediaMetadata({
+          title: toValue(bookName),
+          album: toValue(bookChapterName) || toValue(bookName),
+          artist: toValue(bookAuthorName),
+          artwork: bookCoverSrc
```
```suggestion
          artwork: toValue(bookCoverSrc)
```

pages/reader/epub.vue
williamchong: It is defined below this composable call, so I would have either move their ordering or use another computed :0)
Please advice if moving ordering or keep computed looks better

nwingt: prefer reordering

williamchong: updated

nwingt (approved):

PR #187, opened 2025-07-11
📈 Allow defining UTM in openPaywallModal

williamchong:


composables/use-text-to-speech.ts
```diff
@@ -212,7 +212,11 @@ export function useTextToSpeech(options: TTSOptions = {}) {
 
   async function startTextToSpeech() {
     if (!user.value?.isLikerPlus) {
-      subscription.openPaywallModal()
+      subscription.openPaywallModal({
+        utmSource: 'epub_reader',
+        utmCampaign: nftClassId,
+        utmMedium: 'tts',
+      })
```
nwingt, line 219: I suggest to merge `utmOverrides` with `props` anyway, it is more intuitive to use, no need to an pass empty object

composables/use-subscription.ts
```diff
@@ -102,10 +125,12 @@ export function useSubscription() {
       }
       useLogEvent('begin_checkout', eventPayload.value)
 
+      const analyticsParams = getAnalyticsParameters()
       const { url } = await fetchLikerPlusCheckoutLink({
         period: selectedPlan.value,
         from: getRouteQuery('from'),
-        ...getAnalyticsParameters(),
+        ...analyticsParams,
+        ...utmOverrides,
```
nwingt: Why dont use utm from props?

composables/use-subscription.ts
```diff
@@ -57,14 +57,33 @@ export function useSubscription() {
...
+    if (props.utmMedium) {
+      utmOverrides.utmMedium = props.utmMedium
+    }
+    if (props.utmSource) {
+      utmOverrides.utmSource = props.utmSource
+    }
     return paywallModal.open({
       ...getPaywallModalProps(),
       ...props,
+      onSubscribe: () => startSubscription(utmOverrides),
```
nwingt: if u use utm from props, no need to add this
https://github.com/likecoin/3ook-com/blob/34e057abb0dfa7061244928e69bdd64dcf8d4b08/composables/use-subscription.ts#L51

composables/use-subscription.ts
williamchong: you mean storing utm into props?

nwingt: why not

composables/use-subscription.ts
```diff
@@ -62,6 +66,10 @@ export function useSubscription() {
       paywallModal.close()
     }
 
+    utmCampaign.value = props.utmCampaign
+    utmMedium.value = props.utmMedium
+    utmSource.value = props.utmSource
+
```
nwingt: Image

williamchong: How come we aren't using the existing`selectedPlan` in `@subscribe`..

nwingt (approved):

PR #203, opened 2025-07-17
[TAS-4203] 🚸 Improve opening billing portal

nwingt:


williamchong (commented):
While this can be merged if it works, it seems simpler to make a /plus/manage page with back button and a loading screen, and let user control the timing for canceling?

PR #205, opened 2025-07-17
[TAS-4208] 🐛 Fix navigation handling by using onSelect

AuroraHuang22:


components/BookshelfItem.vue
```diff
@@ -49,9 +49,6 @@
                 class="cursor-pointer"
                 :icon="item.icon"
                 :label="item.label"
-                :href="item.href"
```
williamchong, line 52: Can we set both href and onSelect, for seo internal link reason we do want native href

nwingt:
Please also add pb-safe in the modal for iPhone

image

components/BookshelfItem.vue
```diff
@@ -49,9 +49,6 @@
...
-                :target="item.target"
-                :rel="item.rel"
```
nwingt, line 54: should use `:to="item.to"`

nwingt (changes requested):

nwingt (approved):

PR #221, opened 2025-07-25
[TAS-4277] 🚸 Store TTS options in local storage

williamchong:


nwingt (approved):
Maybe merge #222 first and include key in `accountStore.clearCache()`

stores/account.ts
```diff
@@ -444,9 +444,16 @@ export const useAccountStore = defineStore('account', () => {
       if (!window.localStorage) return
 
       bookKeys.forEach((key) => {
-        getReaderCacheKeySuffixes().forEach((suffix) => {
-          window.localStorage.removeItem(`${key}-${suffix}`)
-        })
+        if (key.startsWith(READER_CACHE_KEY)) {
```
nwingt: isnt book key started with `config.public.cacheKeyPrefix`?

williamchong: Changed to `key.split('-').includes(READER_CACHE_KEY)`

utils/tts.ts
```diff
@@ -25,3 +25,20 @@ export function splitTextIntoSegments(text: string): string[] {
...
+
+export const TTS_CONFIG_KEY = 'tts-config'
+
+const TTS_CONFIG_KEY_SUFFIX_LIST = [
+  'cfi',
+  'dual-page-mode',
+  'locations',
+  'right-to-left',
+  'scale',
+  'tts-index',
```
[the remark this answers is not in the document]
nwingt: @williamchong Seems TTS has these keys only

stores/account.ts
```diff
@@ -444,9 +444,16 @@ export const useAccountStore = defineStore('account', () => {
...
       bookKeys.forEach((key) => {
-        getReaderCacheKeySuffixes().forEach((suffix) => {
-          window.localStorage.removeItem(`${key}-${suffix}`)
-        })
+        if (key.split('-').includes(READER_CACHE_KEY)) {
+          getReaderCacheKeySuffixes().forEach((suffix) => {
+            window.localStorage.removeItem(`${key}-${suffix}`)
+          })
+        }
+        else if (key.split('-').includes(TTS_CONFIG_KEY)) {
```
nwingt: Should be `if`?
```suggestion
        if (key.split('-').includes(TTS_CONFIG_KEY)) {
```

williamchong: but if it has `READER_CACHE_KEY` it won't be under `TTS_CONFIG_KEY`

nwingt (approved):

PR #226, opened 2025-07-28
[TAS-4281] ✨ Support swipe for EPUB reader

nwingt:


pages/reader/epub.vue
```diff
@@ -555,12 +559,76 @@ function onClickTTSPlay() {
...
+      switch (direction) {
+        case 'left':
+          turnPageRight()
+          useLogEvent('reader_navigate_swipe_horizontal')
+          break
+        case 'right':
+          turnPageLeft()
+          useLogEvent('reader_navigate_swipe_horizontal')
+          break
+        case 'up':
```
williamchong, line 618: Dont turn page on swipe up down to prevent accidentally changing page when scrolling?
Also what if i say zoomed in the page?

nwingt: we dont have scrolling/zoom in EPUB now

williamchong (approved):

PR #231, opened 2025-07-29
[TAS-4251] ✨ Add TTS trial quota

williamchong:


server/utils/firestore.ts
```diff
@@ -0,0 +1,35 @@
...
+function initializeFirebase() {
+  if (getApps().length === 0) {
+    app = initializeApp()
+  }
+  else {
+    app = getApps()[0]
+  }
+
+  if (!app) {
+    throw new Error('Failed to initialize Firebase app')
```
nwingt: Prefer using `createError`
```suggestion
    throw createError({ statusMessage: 'Failed to initialize Firebase app' })
```

server/utils/api/user.ts
```diff
@@ -0,0 +1,25 @@
+import { FieldValue } from 'firebase-admin/firestore'
+import { getUserCollection } from '../firestore'
```
nwingt: this should be auto-imported

components/TTSPlayerModal.vue
```diff
@@ -250,7 +254,24 @@ const {
...
+          return
+        }
+      }
+    }
+    toast.add({
+      title: 'Error',
+      description: error.toString(),
+      duration: 5000,
+    })
     console.error('TTS Error:', error)
```
nwingt: Can use `handleError()`

server/api/reader/tts.get.ts
```diff
@@ -1,3 +1,6 @@
+import { isTTSAvailable } from '~/server/utils/api/tts'
+import { getUserDoc } from '~/server/utils/api/user'
+
```
nwingt: Should be auto-imported?

server/utils/api/tts.ts
nwingt, line 1: I prefer putting this file under `/server/utils` so that functions can be auto-imported
Screenshot 2025-07-31 at 23 46 41
Screenshot 2025-07-31 at 23 47 55

server/utils/api/tts.ts
```diff
@@ -0,0 +1,10 @@
+import type { UserDocData } from './user'
+import type { UserSession } from '#auth-utils'
+
+export function isTTSAvailable(session: UserSession, userDoc: UserDocData | undefined): boolean {
+  if (!session || !session.user) return false
+  const isLikerPlus = session.user.isLikerPlus || false
+  if (isLikerPlus) return true
```
nwingt: I want to simplify the input params and follow h3/nuxt-auth utils
```suggestion
import type { H3Event } from 'h3'

export async function getUserTTSAvailable(event: H3Event): Promise<boolean> {
  const session = await getUserSession(event)
  if (!session || !session.user) return false
  const isLikerPlus = session.user.isLikerPlus || false
  if (isLikerPlus) return true
  const userDoc = await getUserDoc(session.user.evmWallet)
```



Caller
```ts
const isTTSAvailable = await getUserTTSAvailable(event)
```

server/utils/api/tts.ts
williamchong: renamed to `api-*.ts`, may suggest better naming

components/TTSPlayerModal.vue
```diff
@@ -250,8 +254,20 @@ const {
   bookAuthorName: props.bookAuthorName,
   bookCoverSrc: props.bookCoverSrc,
   bookLanguage: props.bookLanguage,
-  onError: (error: Event) => {
-    console.error('TTS Error:', error)
+  onError: (error: string | Event | MediaError) => {
+    if (error instanceof MediaError) {
+      if (error.code === MediaError.MEDIA_ERR_SRC_NOT_SUPPORTED) {
+        if (!user.value?.isLikerPlus) {
```
nwingt: Can combine 3 ifs?

nwingt (approved):

PR #237, opened 2025-08-01
[TAS-4329] 🚸 Show different messages from login error

nwingt:


williamchong (approved):
not sure if want more specific error than `REGISTER_INVALID_BODY`

PR #243, opened 2025-08-04
[TAS-4244] 🚸 Add screen loading UI when changing page

williamchong:


pages/reader/epub.vue
```diff
@@ -141,6 +141,19 @@
               ref="reader"
               class="absolute inset-0"
             />
+
+            <!-- Page navigation loading indicator -->
+            <div
+              class="absolute inset-0 flex items-center justify-center z-10 transition-opacity duration-300 opacity-0"
```
nwingt: without `pointer-events-none`, this will block the text selection
I also have some UI adjustments, will open a PR to your branch

nwingt (changes requested):

pages/reader/epub.vue
nwingt: https://github.com/williamchong/liker-land-v3/pull/3

nwingt (approved):

PR #261, opened 2025-08-13
[TAS-4278] ✨ Add zh locale detection for TTS

williamchong:


nwingt (approved):
Please check for any SSR/DOM mismatches

williamchong:
> Please check for any SSR/DOM mismatches

should have no way to trigger SSR for TTS UI currently, but later may want to guard something with client only if needed

PR #264, opened 2025-08-13
📈 Set external_id in fbq

williamchong:


composables/use-logger.ts
```diff
@@ -104,6 +104,19 @@ export function useSetLogUser(user: User | null) {
     console.error('Failed to set user ID in Google Analytics', error)
   }
 
+  const { proxy } = useScriptMetaPixel()
+  if (useRuntimeConfig().public.scripts.metaPixel.id && user?.evmWallet) {
+    try {
+      proxy.fbq('init', useRuntimeConfig().public.scripts.metaPixel.id, {
```
nwingt: perhaps refactor `metaPixelId`?

PR #274, opened 2025-08-15
[TAS-4398] 🥅 Move cart status fetch to client side and retry on 404 in claim

williamchong:
Require https://github.com/likecoin/3ook-com/pull/280

### DIFF
--- a/pages/store/claim.vue
+++ b/pages/store/claim.vue
@@ -33,7 +33,7 @@
         />
       </template>
       <template
-        v-else-if="!isAutoDeliver"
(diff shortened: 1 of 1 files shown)


pages/store/claim.vue
```diff
@@ -170,16 +182,22 @@ await callOnce(nftClassId.value, async () => {
       fatal: true,
     })
   }
+  isLoading.value = false
+  startClaimFlow()
```
nwingt: Should guard with `hasLoggedIn`?

nwingt (approved):

PR #275, opened 2025-08-15
[TAS-4400] 🎨 Clean up Upsell Plus modal code

nwingt:


pages/plus/success.vue
```diff
@@ -132,7 +132,7 @@ onMounted(async () => {
           payment_id: giftPaymentId,
           claiming_token: giftClaimToken,
           cart_id: giftCartId,
-          class_id: giftNFTClassId,
+          nft_class_id: giftNFTClassId,
```
nwingt, line 135: @williamchong I dont see claim page is using this though

williamchong: no not yet, planed to implement class id based claim for legacy purchase

williamchong (approved):

PR #286, opened 2025-08-19
💩 Temp disable monthly upsell to yearly

williamchong:


composables/use-subscription.ts
```diff
@@ -105,8 +105,8 @@ export function useSubscription() {
     }
   }
 
-  async function redirectIfSubscribed() {
-    if (isLikerPlus.value) {
+  async function redirectIfSubscribed(plan?: string) {
+    if (isLikerPlus.value && (!plan || likerPlusPeriod.value !== 'month' || plan !== 'yearly')) {
```
nwingt, line 109: What is this trying to do?

williamchong: To allow monthly to yearly upgrade to pass this check

nwingt (approved):

PR #287, opened 2025-08-19
[TAS-4372] 🎨 Format price to keep two decimal places

AuroraHuang22:


nwingt (commented):
Please add `toFixed(2)` for showing trailing 0 for standard and discounted price

williamchong (approved):

composables/use-subscription.ts
```diff
@@ -177,7 +177,7 @@ export function useSubscription() {
 
   function getPlusDiscountPrice(price: number): number | null {
     if (isLikerPlus.value && price > 0) {
-      return Math.round(price * (1 - PLUS_DISCOUNT_PERCENTAGE) * 100) / 100
+      return Number.parseFloat((price * (1 - PLUS_DISCOUNT_PERCENTAGE)).toFixed(2))
```
williamchong: keep `Math.round` since input is number

williamchong (changes requested):

nwingt (approved):

PR #289, opened 2025-08-19
[TAS-4311] ✨ Add descriptionFull field to bookstoreInfo

AuroraHuang22:


nwingt (approved):

composables/use-book-info.ts
```diff
@@ -31,6 +31,10 @@ export default function (
     return ''
   })
 
+  const description = computed(() => {
```
nwingt, line 34: product page is using this?

PR #294, opened 2025-08-20
👔 Logo points to pricing

williamchong:


nwingt:
Seem only pricing page still has a header, should close this PR?

PR #314, opened 2025-08-26
[TAS-4438] ✨ Support month to year plus API

williamchong:


composables/use-likecoin-session-api.ts
```diff
@@ -262,6 +262,25 @@ export function useLikeCoinSessionAPI() {
     })
   }
 
+  function fetchLikerPlusPriceUpdate({
```
nwingt:
```suggestion
  function updateLikerPlusSubscription({
```

nwingt (approved):

PR #324, opened 2025-08-29
🐛 Fix TTS index out of bound

williamchong:


nwingt (approved):

composables/use-text-to-speech.ts
```diff
@@ -233,7 +233,7 @@ export function useTextToSpeech(options: TTSOptions = {}) {
           if (isTextToSpeechOn.value && isTextToSpeechPlaying.value) {
             playNextElement()
           }
-        }, 500)
+        }, 1000)
```
nwingt: Maybe add a comment for 1000ms?

PR #340, opened 2025-09-05
[TAS-4486] 🐛 Update sold out availability logic

AuroraHuang22:
based on https://github.com/likecoin/nft-book-press/pull/400


williamchong:
Better change `isSoldOut` logic in API instead

PR #344, opened 2025-09-07
🔒️ Add CORS for store GET API

williamchong:


nuxt.config.ts
```diff
@@ -104,6 +104,29 @@ export default defineNuxtConfig({
     },
   },
 
+  routeRules: {
+    '/api/store/products': {
+      security: {
+        corsHandler: {
+          origin: '^https?:\\/\\/([a-zA-Z0-9-]+\\.)?3ook\\.com$',
```
nwingt, line 111: no need for double backslashes in string?

williamchong: seems this string is directly passed into `new RegExp`

PR #348, opened 2025-09-10
[TAS-4536] 📈 Add internal medium and source for recommendation items

williamchong:
Not sure if passing all the way with prop is good idea

### DIFF
--- a/composables/use-book-info.ts
+++ b/composables/use-book-info.ts
@@ -207,6 +207,24 @@ export default function (
     params: { nftClassId: toValue(nftClassId) },
   }))
 
+  function getProductPageRoute(options?: { llMedium?: string, llSource?: string }) {
+    const query: Record<string, string> = {}
+
+    if (options?.llMedium) {
+      query.ll_medium = options.llMedium
+    }
+
+    if (options?.llSource) {
+      query.ll_source = options.llSource
+    }
+
+    return localeRoute({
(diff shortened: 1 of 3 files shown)


pages/store/[nftClassId]/index.vue
```diff
@@ -332,6 +332,8 @@
           :class="getGridItemClassesByIndex(index)"
           :nft-class-id="classId"
           :lazy="true"
+          :ll-medium="'recommendation'"
```
Comment (nwingt) [inline pages/store/[nftClassId]/index.vue:341]:
```suggestion
          ll-medium="recommendation"
```

composables/use-book-info.ts
```diff
@@ -202,10 +210,23 @@ export default function (
...
+    if (llSource && toValue(llSource)) {
+      query.ll_source = toValue(llSource)
+    }
+
+    return localeRoute({
+      name: 'store-nftClassId',
+      params: { nftClassId: toValue(nftClassId) },
+      query: Object.keys(query).length > 0 ? query : undefined,
+    })
+  })
```
nwingt: maybe add a function `getProductPageRoute({ llMedium, llSource })` instead of injecting book info composable?

nwingt (commented):
m

nwingt (approved):

nwingt:
Not sure if we should reuse `getProductPageRoute()` in computed `productPageRoute`

PR #352, opened 2025-09-11
[TAS-4503] 💄 Display voice avatar on TTS button

williamchong:
image

### DIFF
--- a/composables/use-tts-voice.ts
+++ b/composables/use-tts-voice.ts
@@ -1,0 +1,115 @@
+import { useStorage } from '@vueuse/core'
+import phoebeAvatar from '@/assets/images/voice-avatars/phoebe.jpg'
+import leiTingYinAvatar from '@/assets/images/voice-avatars/lei-ting-yin.jpg'
+import pazuAvatar from '@/assets/images/voice-avatars/pazu.jpg'
+import defaultAvatar from '@/assets/images/voice-avatars/default.jpg'
+
+interface TTSVoiceOptions {
+  bookLanguage?: string | Ref<string> | ComputedRef<string>
+}
+
+export function useTTSVoice(options: TTSVoiceOptions = {}) {
+  const { bookLanguage } = options
+
+  const config = useRuntimeConfig()
+
+  const ttsConfigCacheKey = computed(() =>
+    [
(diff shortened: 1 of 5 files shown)


AuroraHuang22 (approved):

nwingt (commented):
I think it is better to move `ttsLanguageVoice`, `activeTTSLanguageVoiceAvatar`,etc, to another composable, e.g. `useTTSVoice()`, so that we dont have to include a huge composable `useTextToSpeech()` just for displaying an avatar

composables/use-tts-voice.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+    return ttsLanguageVoiceOptions
+  })
+
+  const ttsLanguageVoiceValues = availableTTSLanguageVoiceOptions.value.map(option => option.value)
+
+  const ttsPlaybackRateOptions = computed(() => [0.5, 0.75, 1.0, 1.25, 1.5].map(rate => ({
+    label: rate === 1.0 ? $t('reader_text_to_speech_normal_speed') : `${rate}x`,
+    value: rate,
+  })))
```
nwingt: `ttsPlaybackRateOptions` should be stayed in `useTextToSpeech`?

composables/use-tts-voice.ts
```diff
@@ -0,0 +1,124 @@
...
+      else if (locales.some(locale => locale.startsWith('zh') && !locale.includes('-hk'))) {
+        voice = ttsLanguageVoiceValues.find(value => value.startsWith('zh-TW')) || voice
+      }
+    }
+
+    return voice
+  }
+
+  const ttsLanguageVoice = useStorage(getTTSConfigKeyWithSuffix(ttsConfigCacheKey.value, 'voice'), getDefaultTTSVoiceByLocale())
+  const ttsPlaybackRate = useStorage(getTTSConfigKeyWithSuffix(ttsConfigCacheKey.value, 'playback-rate'), 1.0)
```
nwingt: Same for `ttsPlaybackRate`

composables/use-tts-voice.ts
williamchong: ai dumb

nwingt (approved):

PR #361, opened 2025-09-17
♻️ Remove owner from nft store, use bookshelf as truth

williamchong:
`ownerInfo` from aggregated info is unreliable, will refactor in API later


pages/store/claim.vue
```diff
@@ -239,9 +239,10 @@ async function checkItemsDeliveryThroughIndexer() {
   // TODO: Handle multiple items
   await Promise.all([
```
nwingt: can remove `Promise.all`?

nwingt (approved):

PR #362, opened 2025-09-18
♻️ Refactor image util to composable

williamchong:


utils/index.ts
```diff
@@ -1,25 +1,7 @@
...
-
-  const config = useRuntimeConfig()
-  const params = new URLSearchParams()
-  params.set('url', imageURL)
-  if (size) params.set('width', size.toString())
-  return `${config.public.likeCoinStaticEndpoint}/thumbnail/?${params.toString()}`
-}
-
 export function getResizedNormalizedImageURL(imageURL: string, { size }: { size?: number } = {}) {
+  const { normalizeURIToHTTP, getResizedImageURL } = useCoverImage()
```
nwingt: since it uses `useCoverImage()`, should move out from `/utils` too?

PR #364, opened 2025-09-18
🏷️ Add type and indexer functions

williamchong:


types/nft.d.ts
```diff
@@ -35,42 +42,42 @@ declare interface NFTClassMetadata {
...
+  owner?: NFTAccount
 }
 
 declare type NFTAttributeValue = string | number | { [key: string]: string | number }
 
 declare interface NFTAttribute {
   trait_type: string
-  value: NFTAttributeValue
-  display_type?: string
+  value: string
```
nwingt: then `NFTAttributeValue` can be removed too?

shared/utils/indexer.ts
```diff
@@ -3,20 +3,50 @@ export interface PaginationInfo {
   count: number
 }
 
+export interface PaginationResponse {
+  next_key?: number
+  count: number
+}
+
+export interface Error {
```
nwingt: Will this shadow native Error?

types/nft.d.ts
williamchong: lets just use `NFTAttributeValue`

shared/utils/indexer.ts
williamchong: Seems acutally no used in any function, will remove for now

nwingt (approved):

PR #369, opened 2025-09-26
[TAS-4538] ✨ Add TTS samples to PaywallModal

nwingt:
localhost_3000_pricing (2)

Screenshot 2025-09-30 at 00 42 28


composables/use-tts-samples-player.ts
```diff
@@ -0,0 +1,235 @@
...
+      onError?.(new Error('Audio source not available'))
+      return
+    }
+
+    // Set up new playback
+    currentSegmentIndex.value = 0
+    playCurrentSegment()
+  }
+
+  function cleanUpAudio() {
```
williamchong: Seems `resetAudio` is more suitable than the name `cleanUp`

composables/use-tts-samples-player.ts
```diff
@@ -0,0 +1,235 @@
...
+  async function playCurrentSegment() {
+    const segment = currentSegment.value
+    if (!segment) return
+
+    try {
+      const audioElement = createAudio(segment)
+
+      // Wait for audio to be ready before playing
+      await new Promise<void>((resolve, reject) => {
+        audioElement.oncanplaythrough = () => {
```
williamchong: What happen if we don't wait for `oncanplaythrough`?

williamchong (approved):

PR #373, opened 2025-09-30
💄 Update plus CTA button to show trial

williamchong:


components/UpsellPlusModal.vue
```diff
@@ -128,17 +128,21 @@
         >
           <span
             class="font-bold"
-            v-text="$t('upsell_plus_yearly_button')"
+            v-text="props.nftClassId
+              ? $t('upsell_plus_yearly_gift_cta')
+              : (props.trialPeriodDays && allowYearlyTrial
+                ? $t('upsell_plus_yearly_trial_cta', { days: props.trialPeriodDays })
+                : $t('upsell_plus_yearly_button'))"
```
AuroraHuang22: Might be cleaner to extract this into a computed property

nwingt: Agree

nwingt (approved):

components/UpsellPlusModal.vue
```diff
@@ -224,9 +222,30 @@ const showYearlyPlan = computed(() => (
...
+  if (props.nftClassId) {
+    return $t('upsell_plus_yearly_gift_cta')
+  }
+  if (props.trialPeriodDays && allowYearlyTrial.value) {
+    return $t('upsell_plus_yearly_trial_cta', { days: props.trialPeriodDays })
+  }
+  return $t('upsell_plus_yearly_button')
+})
+
+const monthlyButtonCta = computed(() => {
```
nwingt:
```suggestion
const yearlyButtonCTA = computed(() => {
  if (props.nftClassId) {
    return $t('upsell_plus_yearly_gift_cta')
  }
  if (props.trialPeriodDays && allowYearlyTrial.value) {
    return $t('upsell_plus_yearly_trial_cta', { days: props.trialPeriodDays })
  }
  return $t('upsell_plus_yearly_button')
})

const monthlyButtonCTA = computed(() => {
```

nwingt (changes requested):

PR #375, opened 2025-09-30
[TAS-4643] ✨ Add search function to store page

williamchong:
image


server/utils/airtable.ts
```diff
@@ -57,13 +57,75 @@ export interface FetchAirtableCMSProductsByTagIdResponseData {
       'Min Price'?: number
       'Listing Date'?: string[]
       'Timestamp'?: number
-      'DRM-free'?: number // 0 = false, 1 = true
+      'DRM-free'?: boolean | number // boolean in Publications table, number (0 = false, 1 = true) in Products table
```
nwingt, line 60: i can fix the type to boolean in Airtable

nwingt (approved):

server/utils/airtable.ts
williamchong: Lets do it when you add search box

PR #380, opened 2025-10-03
[TAS-4656] ✨ Add maintenance page

nwingt:
Screenshot 2025-10-03 at 12 54 14


components/AppLogo.vue
```diff
@@ -18,9 +50,9 @@ const props = defineProps({
     type: [Number, String],
     default: 20,
   },
-  fill: {
-    type: String,
-    default: '#131313',
+  isPadded: {
+    type: Boolean,
+    default: true,
   },
```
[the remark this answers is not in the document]
nwingt: dont want to change fill color

AuroraHuang22 (approved):

.env.example
```diff
@@ -10,6 +10,7 @@ AZURE_SUBSCRIPTION_KEY=
 BASE_URL=http://localhost:3000
 GA_TRACKING_ID=
 IPFS_ENDPOINT=
+IS_MAINTENANCE_MODE=FALSE
```
williamchong:
```suggestion
IS_MAINTENANCE_MODE=
```
I think not setting env means `FALSE` wouldbe more robust
Otherwise we might accidentally get stuff like `True` `TRUE` `1` `False` `false` `0` etc

apphosting.mainnet.yaml
```diff
@@ -39,6 +39,8 @@ env:
     value: southeastasia
   - variable: AZURE_SUBSCRIPTION_KEY
     secret: AZURE_SUBSCRIPTION_KEY
+  - variable: IS_MAINTENANCE_MODE
+    value: FALSE
```
williamchong:
```suggestion
  - variable: IS_MAINTENANCE_MODE
    value:
```
Or just remove the key if yaml complains

nuxt.config.ts
```diff
@@ -76,6 +76,7 @@ export default defineNuxtConfig({
       walletConnectProjectId: process.env.WALLET_CONNECT_PROJECT_ID,
       arweaveEndpoint: process.env.ARWEAVE_ENDPOINT,
       ipfsEndpoint: process.env.IPFS_ENDPOINT,
+      isMaintenanceMode: process.env.IS_MAINTENANCE_MODE === 'TRUE',
```
williamchong:
```suggestion
      isMaintenanceMode: process.env.IS_MAINTENANCE_MODE,
```

williamchong (changes requested):

nuxt.config.ts
nwingt: Updated with `!== undefined` for boolean type

williamchong (approved):

PR #398, opened 2025-10-21
[TAS-4454] 💄 Improve account page UI

nwingt:


pages/account.vue
```diff
@@ -294,4 +372,58 @@ async function handleClearReaderCacheButtonClick() {
...
+      duration: 3000,
+      color: 'error',
+    })
+  }
+}
+
+async function handleEVMWalletClick() {
+  useLogEvent('account_evm_wallet_click')
+  try {
+    await navigator.clipboard.writeText(user.value?.evmWallet || '')
```
williamchong: Should we use `useClipboard` instead?

pages/account.vue
```diff
@@ -294,4 +372,58 @@ async function handleClearReaderCacheButtonClick() {
     await handleError(error, { title: $t('account_page_reader_cache_clear_failed') })
   }
 }
+
+async function handleLikerIdClick() {
+  useLogEvent('liker_id_wallet_click')
+  try {
+    await navigator.clipboard.writeText(user.value?.likerId || '')
```
williamchong: Should we use `useClipboard` instead?

pages/account.vue
```diff
@@ -63,71 +126,75 @@
...
+                variant="ghost"
+                color="neutral"
+                size="xs"
+                @click="handleEVMWalletClick"
               />
+            </UTooltip>
+
+            <template #right>
               <UButton
-                v-if="user?.loginMethod === 'magic'"
```
williamchong, line 76: Why can we remove this guard?

i18n/locales/zh-Hant.json
```diff
@@ -26,11 +26,14 @@
...
-  "account_page_account_title": "個人帳戶",
+  "account_page_account_display_name": "顯示名稱",
+  "account_page_account_id": "推薦碼",
+  "account_page_account_id_tooltip": "不對外公開",
+  "account_page_account_title": "帳戶",
   "account_page_book_list": "書單",
   "account_page_contact_support": "客戶服務",
-  "account_page_cosmos_wallet": "錢包地址 (LikeCoin)",
+  "account_page_cosmos_wallet": "舊版錢包",
+  "account_page_display_name_tooltip": "用於向人推薦書目、作者簽名扉頁的上款等。",
```
williamchong, line 35: This description would be invalid for those with actual display name set?

i18n/locales/en.json
```diff
@@ -26,27 +26,29 @@
...
-  "account_page_account_title": "User Account",
+  "account_page_account_display_name": "Display Name",
+  "account_page_account_id": "Referral Code",
+  "account_page_account_id_tooltip": "Not publicly disclosed",
+  "account_page_account_title": "Account",
   "account_page_book_list": "3ookList",
   "account_page_contact_support": "Customer service",
-  "account_page_cosmos_wallet": "Wallet Address (LikeCoin)",
+  "account_page_cosmos_wallet": "Legacy Wallet",
+  "account_page_display_name_tooltip": "Used for referring books to people, the recipient's name on inscription pages signed by authors, etc.",
```
williamchong, line 35: This description would be invalid for those with actual display name set?

pages/account.vue
nwingt: updated

i18n/locales/zh-Hant.json
nwingt: @ckxpress I think `向人推薦書目` misled William

pages/account.vue
nwingt: Updated

pages/account.vue
nwingt: Updated

williamchong (approved):
let’s just try this version for now

PR #401, opened 2025-10-21
[TAS-4375] ➕ Add posthog

williamchong:
We mainly want the experiment feature, and maybe feature flags
Lets integrate event and id before we start any experiment


composables/use-logger.ts
```diff
@@ -160,4 +171,18 @@ export function useSetLogUser(user: User | null) {
       console.error('Failed to set user data in Intercom', error)
     }
   }
+
+  const { $posthog } = useNuxtApp()
```
nwingt, line 175: I think we should call `useNuxtApp` at the function top

nwingt (approved):

PR #406, opened 2025-10-23
[TAS-4626] 💩 Disable double buffer and detect stalled TTS audio

williamchong:


composables/use-text-to-speech.ts
```diff
@@ -172,15 +173,27 @@ export function useTextToSpeech(options: TTSOptions = {}) {
...
-        console.warn('Audio playback error:', error)
-        options.onError?.(error)
-        setTimeout(() => {
-          if (isTextToSpeechOn.value && isTextToSpeechPlaying.value) {
-            playNextElement()
+      audio.onstalled = () => {
+        if (bufferIndex === currentBufferIndex.value && audio) {
+          console.warn(`Audio playback stalled at ${ttsPlaybackRate.value}x for text: "${audio.getAttribute('data-text')}"`)
+          if (audio.currentTime < 0.00001) {
+            // safari on iOS sometimes gets stuck at 0.000001 for rate > 1.0
```
```suggestion
            // Safari on iOS sometimes gets stuck at 0.000001 for rate > 1.0
```

nwingt (commented):
is the PR title correct? should be re-enabled double buffer for iOS PWA?

williamchong:
> is the PR title correct? should be re-enabled double buffer for iOS PWA?

It is correct. Now we dont call
```
currentBufferIndex.value = idleBufferIndex.value
```
to swap active audio, so it is always the same audio element being used as active. Now idle audio element is used as preloading only.

PR #409, opened 2025-10-24
[TAS-4716] 👽 Support utm_term for pricing page campaign ID

nwingt:


composables/use-pricing-page-campaign.ts
```diff
@@ -15,8 +15,15 @@ const CAMPAIGNS: Record<string, PricingPageCampaign> = {
...
 
+const NORMALIZED_CAMPAIGN_IDS: Record<string, string> = {
+  '120230946283520669': 'fragmentedtime',
+}
+
 export function usePricingPageCampaign(options: { campaignId: MaybeRefOrGetter<string | undefined> }) {
-  const campaignId = computed(() => toValue(options.campaignId))
+  const campaignId = computed(() => {
+    const id = toValue(options.campaignId)
+    return NORMALIZED_CAMPAIGN_IDS[id || ''] || id
```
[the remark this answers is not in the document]
nwingt: undefined is still undefined, do you know what you are talking about?

composables/use-pricing-page-campaign.ts
```diff
@@ -15,8 +15,15 @@ const CAMPAIGNS: Record<string, PricingPageCampaign> = {
...
+  })
```
williamchong: Please change to use `utm_term` since we have multiple `utm_campaign` (campaign) and `utm_content` (adset) that might use the same theme
We would manually assign `utm_term` from ad landing url to target which theme to show

williamchong (changes requested):

composables/use-pricing-page-campaign.ts
nwingt: Updated

PR #415, opened 2025-10-27
⚗️ Add ab test for pricing campaign

williamchong:


nwingt (approved):

components/PaywallModal.vue
```diff
@@ -331,7 +331,7 @@ const utmCampaign = computed(() => {
   return getRouteQuery('utm_campaign') || props.utmCampaign
 })
 const campaignId = computed(() => {
-  return getRouteQuery('utm_term') || utmCampaign.value
+  return getRouteQuery('utm_term') || getRouteQuery('utm_campaign')
```
nwingt, line 334: Since we have
https://github.com/likecoin/3ook-com/blob/8a16ac22e8311dd1ef989b10e1e147b45ff5af30/pages/pricing.vue#L145-L155

How about?
```suggestion
  return getRouteQuery('utm_term') || getRouteQuery('utm_campaign') || utmCampaign.value
```

williamchong: But `utmCampaign.value` would default to `pricing_page`, which is an invalid campaign id

nwingt (approved):

PR #431, opened 2025-10-31
🐛 Also show staked non-owned book in /shelf

williamchong:


ckxpress:
is it possible that the user will be able to open a staked but not purchased book？

williamchong:
> is it possible that the user will be able to open a staked but not purchased book？

No but good catch, should block in UI

nwingt:
Suggest to update variable naming
https://github.com/williamchong/liker-land-v3/pull/10

PR #435, opened 2025-11-01
👔 Show staking donate button only in testnet

williamchong:


nwingt (commented):
How about just showing the button when `config.public.isTestnet` is true?

nwingt (approved):

PR #438, opened 2025-11-02
[TAS-4816] 💬 Update deposit page text

nwingt:


i18n/locales/en.json
```diff
@@ -103,6 +104,7 @@
   "bookshelf_please_login": "Please login to view your 3ookShelf",
   "bookshelf_view_book_product_page": "View in 3ookStore",
   "bookshelf_view_book_staking_page": "Stake on this Book",
+  "breadcrumb_account_deposit": "帳戶 ＞ LikeCoin Deposit",
```
```suggestion
  "breadcrumb_account_deposit": "Account ＞ LikeCoin Deposit",
```
Should be english?

nwingt: Updated

williamchong (approved):

PR #439, opened 2025-11-03
🏗️ Restructure account pages for breadcrumb & add events

nwingt:
Please ignore whitespaces to review

Screenshot 2025-11-03 at 11 40 06


layouts/default.vue
```diff
@@ -1,18 +1,22 @@
 <template>
-  <div class="flex flex-col min-h-svh pb-17 mb-(--intercom-launcher-offset)">
+  <div class="flex flex-col min-h-svh pb-(--intercom-launcher-offset)">
```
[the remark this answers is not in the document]
nwingt: Intended

pages/account.vue
```diff
@@ -0,0 +1,118 @@
+<template>
+  <NuxtLayout
+    name="default"
+    class="relative gap-4 w-full max-w-(--breakpoint-phone) mx-auto"
```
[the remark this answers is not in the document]
nwingt: layout is flex

pages/account/deposit.vue
```diff
@@ -1,323 +1,285 @@
...
-              />
-              <span
-                class="text-xs font-semibold text-muted"
-                v-text="t('governance_page_staked_balance')"
-              />
-            </div>
-            <BalanceLabel
-              class="text-xl"
-              :value="governanceData.formattedLikeStakedBalance.value"
+      <div class="grid phone:grid-cols-2 gap-4">
```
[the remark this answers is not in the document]
nwingt: default is one

pages/account/deposit.vue
```diff
@@ -1,323 +1,285 @@
...
+            :disabled="!withdrawAmount || isLoading || governanceData.isWithdrawLocked.value"
+            @click="handleWithdraw"
+          />
+        </div>
+      </UCard>
+    </section>
   </main>
+  <!-- Empty State -->
+  <div
+    v-else-if="hasLoggedIn && !walletAddress"
```
williamchong, line 267: Would this be a empty component and explode if it doesn't fall into the 2 `v-if-else-if` cases?
Should we have a fallback `v-else`?

pages/account.vue
```diff
@@ -0,0 +1,118 @@
...
+    >
+      <h1
+        class="text-2xl font-bold"
+        v-text="breadcrumbItems[0]?.label"
+      />
+    </header>
+
+    <!-- Login Prompt -->
+    <aside
+      v-if="!hasLoggedIn"
```
williamchong, line 51: Please confirm this condition is correct and expected to not be a `v-else`

nwingt: header is header and login panel are two separated sections

pages/account/deposit.vue
nwingt: empty component is expected

PR #440, opened 2025-11-03
[TAS-4837] ✨ Show and sort bookstore items by staking

nwingt:
Based on #439

Screenshot 2025-11-03 at 11 26 02

### Bookshelf Before
Screenshot 2025-11-03 at 11 25 09

### Bookshelf after
Screenshot 2025-11-03 at 11 25 22

### DIFF
--- a/pages/store/index.vue
+++ b/pages/store/index.vue
@@ -177,8 +177,8 @@
           :book-name="item.title"
           :book-cover-src="item.imageUrl"
           :price="item.minPrice"
-          :total-staked="isStakingTagId ? Number(formatUnits(item.totalStaked ?? 0n, likeCoinTokenDecimals)) : 0"
-          :staker-count="isStakingTagId ? (item.stakerCount ?? 0) : 0"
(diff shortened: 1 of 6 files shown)


pages/store/index.vue
```diff
@@ -567,7 +594,11 @@ async function fetchItems({ lazy = false, isRefresh = false } = {}) {
   }
 
   try {
-    await bookstoreStore.fetchCMSProductsByTagId(localizedTagId.value, { isRefresh })
+    await Promise.all([
+      bookstoreStore.fetchCMSProductsByTagId(localizedTagId.value, { isRefresh }),
+      // NOTE: Fetch staking books for sorting items
+      bookstoreStore.fetchStakingBooks('total_staked', { isRefresh }).catch(() => {}),
```
williamchong: `fetchStakingBooks` is a paginated API so there is not guarantee it would include all books in a tag, maybe we also need some kind of lazy fetching?
For now maybe we should at least set `fetchStakingBooks` page size to max (100)

PR #457, opened 2025-11-10
💩 Fix USwitch out sync with v-model issue

williamchong:


pages/account/deposit.vue
```diff
@@ -291,10 +291,20 @@ const { balanceOf } = useLikeCoinContract()
...
 
+watch(isAutoRestakeEnabled, (newValue) => {
+  isAutoRestakeEnabledStorage.value = newValue
+})
+
+onMounted(() => {
+  // HACK: Mitigate USwitch initial value out sync wih v-model issue
+  isAutoRestakeEnabled.value = isAutoRestakeEnabledStorage.value
+})
+
```
nwingt, line 307: How about using computed with setter?
```suggestion
// HACK: Mitigate USwitch initial value out sync with v-model issue
const isAutoRestakeEnabled = computed({
  get: () => isAutoRestakeEnabledStorage.value,
  set: (value) => {
    isAutoRestakeEnabledStorage.value = value
  },
})
```

nwingt (approved):

pages/account/deposit.vue
williamchong: Seems a value change needs to be triggered on mounted to prevent SSR issue

PR #472, opened 2025-11-14
[TAS-4453] 💄 Add back to store navigation label for product page

AuroraHuang22:
https://github.com/user-attachments/assets/9c2ccaa3-9fa1-4d3d-9c7d-0c3f3d03bd77

### DIFF
--- a/composables/use-store-page-state.ts
+++ b/composables/use-store-page-state.ts
@@ -1,0 +1,66 @@
+export function useStorePageState() {
+  const localeRoute = useLocaleRoute()
+  const getRouteQuery = useRouteQuery()
+
+  const lastScrollPosition = useState<number>('store-last-scroll-position', () => 0)
+  const lastVisitedTag = useState<string>('store-last-visited-tag', () => '')
+  const lastVisitedQuery = useState<Record<string, string>>('store-last-visited-query', () => ({}))
+
+  function save(tag: string, query: Record<string, string>) {
+    if (import.meta.client) {
+      lastScrollPosition.value = window.scrollY || document.documentElement.scrollTop
+    }
+    lastVisitedTag.value = tag
+    lastVisitedQuery.value = { ...query }
+  }
+
+  function restoreScroll() {
+    if (import.meta.client && lastScrollPosition.value > 0) {
+      window.scrollTo({
+        top: lastScrollPosition.value,
+      })
+    }
+  }
+
+  function clear() {
+    lastScrollPosition.value = 0
+    lastVisitedTag.value = ''
+    lastVisitedQuery.value = {}
+    if (import.meta.client) {
+      window.scrollTo({
+        top: 0,
+      })
+    }
+  }
+
+  async function restoreIfNeeded() {
+    if (!lastVisitedTag.value) {
+      return
+    }
+    await navigateTo(localeRoute({
+      name: 'store',
+      query: {
+        ...lastVisitedQuery.value,
+        tag: lastVisitedTag.value === 'default' ? undefined : lastVisitedTag.value,
+      },
+    }), { replace: true })
+    return
+  }
+
+  function restoreScrollIfNeeded() {
+    const currentTag = getRouteQuery('tag', '')
+    const savedTag = lastVisitedTag.value
+
+    if (currentTag === savedTag && lastScrollPosition.value > 0) {
+      restoreScroll()
+    }
+  }
+
+  return {
+    save,
+    restoreScroll,
+    clear,
+    restoreIfNeeded,
+    restoreScrollIfNeeded,
+  }
+}
--- a/pages/store/[nftClassId]/index.vue
+++ b/pages/store/[nftClassId]/index.vue
@@ -1268,14 +1268,9 @@ function handleKeywordClick(keyword: string) {
 
 async function handleBackButtonClick() {
   useLogEvent('product_page_back_button_click')
-  if (window.history.length > 1) {
-    router.back()
-  }
-  else {
-    await navigateTo(localeRoute({
-      name: 'store',
-      query: route.query,
-    }))
-  }
+  await navigateTo(localeRoute({
+    name: 'store',
+    query: route.query,
+  }))
 }
 </script>
--- a/pages/store/index.vue
+++ b/pages/store/index.vue
@@ -589,12 +589,13 @@ useHead(() => {
 watch(
   () => route.query.tag,
   async (newTag, oldTag) => {
-    if (oldTag && !newTag && !isSearchMode.value) {
-      storePageState.clear()
-    }
     if (newTag !== oldTag) {
       await fetchItems({ lazy: true })
+
       if (!isSearchMode.value) {
+        if (!newTag) {
+          storePageState.clear()
+        }
         storePageState.restoreScrollIfNeeded()
       }
     }
--- a/pages/store/index.vue
+++ b/pages/store/index.vue
@@ -813,8 +813,13 @@ onMounted(async () => {
   ])
 })
 
-onBeforeUnmount(() => {
-  storePageState.save(tagId.value, route.query as Record<string, string>)
+onBeforeRouteLeave((to) => {
+  if (to.name && String(to.name).startsWith('store')) {
+    storePageState.save(tagId.value, route.query as Record<string, string>)
+  }
+  else {
+    storePageState.clear()
+  }
 })
 
 watch(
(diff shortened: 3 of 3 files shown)


nwingt:
@AuroraHuang22 Please rebase first

pages/store/[nftClassId]/index.vue
```diff
@@ -1,5 +1,31 @@
...
+        <UIcon
+          name="i-material-symbols-arrow-back-rounded"
+          class="w-5 h-5 text-gray-400 group-hover:text-gray-600 duration-75"
+        />
+        <p
+          class="text-md text-gray-400 font-semibold group-hover:text-gray-600 duration-75"
+          v-text="$t('product_page_back_to_store_label')"
+        />
+      </button>
+    </div>
```
Comment (nwingt) [inline pages/store/[nftClassId]/index.vue]: Please refer to account page header

pages/store/[nftClassId]/index.vue
```diff
@@ -1166,4 +1187,13 @@ function calculateCustomPrice(editionPrice: number, tippingAmount: number | unde
   const base = Number(editionPrice) || 0
   return Number((tip + base).toFixed(2))
 }
+
+async function handleBackToStore() {
```
Comment (nwingt) [inline pages/store/[nftClassId]/index.vue]:
```suggestion
async function handleBackButtonClick() {
  useLogEvent('product_page_back_button_click')
```

pages/store/index.vue
```diff
@@ -236,6 +236,7 @@ const metadataStore = useMetadataStore()
 const infiniteScrollDetectorElement = useTemplateRef<HTMLLIElement>('infiniteScrollDetector')
 const shouldLoadMore = useElementVisibility(infiniteScrollDetectorElement)
 const { handleError } = useErrorHandler()
+const { restoreScroll, saveState: savePageState, lastVisitedTag, lastVisitedQuery, clearState } = useStorePageState()
```
nwingt: Prefer not spreading properties and methods
```suggestion
const storePageState = useStorePageState()
```

pages/store/index.vue
```diff
@@ -667,7 +668,22 @@ async function fetchItems({ lazy = false, isRefresh = false } = {}) {
...
+    shouldRestoreScrollAfterLoad.value = true
+    await navigateTo(localeRoute({
+      name: 'store',
+      query: {
+        ...lastVisitedQuery.value,
+        tag: lastVisitedTag.value === 'listing' ? undefined : lastVisitedTag.value,
+      },
+    }), { replace: true })
+    return
+  }
```
nwingt: I think can move into `useStorePageState()` and create a new function `restoreIfNeeded()`

pages/store/index.vue
```diff
@@ -706,11 +722,19 @@ onMounted(async () => {
...
+
 watch(
   tag,
   async (tag) => {
     if (tag) {
       await fetchItems({ lazy: true })
+      if (shouldRestoreScrollAfterLoad.value) {
+        restoreScroll()
+        shouldRestoreScrollAfterLoad.value = false
+      }
```
nwingt: create another function `restoreScrollIfNeeded()` in `useStorePageState()`

composables/use-store-page-state.ts
```diff
@@ -0,0 +1,41 @@
+export function useStorePageState() {
+  const lastScrollPosition = useState<number>('store-last-scroll-position', () => 0)
+  const lastVisitedTag = useState<string>('store-last-visited-tag', () => '')
+  const lastVisitedQuery = useState<Record<string, string>>('store-last-visited-query', () => ({}))
+
+  function saveState(tag: string, query: Record<string, string>) {
```
nwingt:
```suggestion
  function save(tag: string, query: Record<string, string>) {
```

composables/use-store-page-state.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  function restoreScroll() {
+    if (import.meta.client && lastScrollPosition.value > 0) {
+      window.scrollTo({
+        top: lastScrollPosition.value,
+      })
+    }
+  }
+
+  function clearState() {
```
nwingt:
```suggestion
  function clear() {
```

composables/use-store-page-state.ts
```diff
@@ -0,0 +1,41 @@
...
+      window.scrollTo({
+        top: 0,
+      })
+    }
+  }
+
+  return {
+    lastScrollPosition,
+    lastVisitedTag,
+    lastVisitedQuery,
```
nwingt: Can remove these if using internally

pages/store/index.vue
```diff
@@ -667,7 +668,22 @@ async function fetchItems({ lazy = false, isRefresh = false } = {}) {
...
+
 onMounted(async () => {
+  const shouldRestoreLastVisited = !route.query.tag && lastVisitedTag.value && !isSearchMode.value
+  if (shouldRestoreLastVisited) {
+    shouldRestoreScrollAfterLoad.value = true
+    await navigateTo(localeRoute({
+      name: 'store',
+      query: {
+        ...lastVisitedQuery.value,
+        tag: lastVisitedTag.value === 'listing' ? undefined : lastVisitedTag.value,
```
williamchong: We shouldn't need to store `tag` and other query into store, since they are already in URL and will not be lost if we use `router.back()`? We probably just need scroll position

pages/store/index.vue
```diff
@@ -706,11 +722,19 @@ onMounted(async () => {
...
+  savePageState(tagId.value, route.query as Record<string, string>)
+})
+
 watch(
   tag,
   async (tag) => {
     if (tag) {
       await fetchItems({ lazy: true })
+      if (shouldRestoreScrollAfterLoad.value) {
+        restoreScroll()
```
williamchong: I think to make it simpler, check here should be sth like `storedTag === currentTag && storedScroll` exist, then we just clear the stored values after using it

williamchong (changes requested):

pages/store/index.vue
```diff
@@ -798,8 +806,21 @@ onMounted(async () => {
   ])
 })
 
+onBeforeUnmount(() => {
+  storePageState.save(tagId.value, route.query as Record<string, string>)
+})
+
+watch(
+  () => route.query.tag,
```
nwingt: is it same as watch tagId?

there is another watcher in L588

pages/store/index.vue
```diff
@@ -798,8 +806,21 @@ onMounted(async () => {
...
+  (newTag, oldTag) => {
+    if (oldTag && !newTag && !isSearchMode.value) {
+      storePageState.clear()
+    }
+  },
+)
+
 watch(
-  shouldLoadMore,
+  () => shouldLoadMore.value,
```
nwingt: isnt `shouldLoadMore` the same as `() => shouldLoadMore.value`?

pages/store/index.vue
```diff
@@ -798,8 +806,21 @@ onMounted(async () => {
   ])
 })
 
+onBeforeUnmount(() => {
```
nwingt: Since the back button only appear on pages under /store 
Maybe use `onBeforeRouteLeave((to, from) => {})` and save state when route name begins with store only?

AuroraHuang22: We also restore state when navigating back to store from /account, so not restricting the source route works better.

nwingt: I would expect users come to store from account see a clean state
in what circumstances we need this behavior?

nwingt: I think the objective of this task is solving the back behavior from product page to store page

AuroraHuang22: [♻️ Improve page state management](https://github.com/likecoin/3ook-com/pull/472/commits/21bc2e98d56158a6f503219594630a8068baaf18)

pages/store/index.vue
```diff
@@ -586,11 +586,20 @@ useHead(() => {
...
+  async (newTag, oldTag) => {
+    if (oldTag && !newTag && !isSearchMode.value) {
+      storePageState.clear()
+    }
+    if (newTag !== oldTag) {
+      await fetchItems({ lazy: true })
+      if (!isSearchMode.value) {
+        storePageState.restoreScrollIfNeeded()
+      }
+    }
```
williamchong:
```suggestion
    if (newTag !== oldTag) {
      await fetchItems({ lazy: true })

      if (!isSearchMode.value) {
        if (!newTag) {
          storePageState.clear()
        }
        storePageState.restoreScrollIfNeeded()
      }
    }
```

williamchong (changes requested):

PR #482, opened 2025-11-19
[TAS-4873] 🚸 Improve deposit UI/UX

nwingt:


components/AmountInputModal.vue
```diff
@@ -70,20 +74,34 @@ const props = withDefaults(defineProps<{
...
   ticker?: string
   confirmButtonTitle?: string
+  isShowBalance?: boolean
 }>(), {
   ticker: 'LIKE',
 })
 
 const emit = defineEmits(['close'])
 
+const { user } = useUserSession()
```
williamchong: Should just pass `wallet` or event `balance` in props instead of using user

composables/use-governance-data.ts
```diff
@@ -89,18 +89,30 @@ export function useGovernanceData(walletAddress: string | Ref<string>) {
...
+      minimumFractionDigits: 2,
+      maximumFractionDigits: 6,
+    })
+  })
+
+  const estimatedRewardAPY = computed(() => {
+    if (!veLikeBalance.value) {
+      return 0
     }
+    return estimatedRewardPerDay.value * 365 / Number(formatUnits(veLikeBalance.value, likeCoinTokenDecimals))
```
williamchong, line 112: Not sure if this is actually `APR` not `APY`

utils/index.ts
```diff
@@ -67,3 +67,16 @@ export function parseURIString<T>(uri: string): T | undefined {
 export function shortenWalletAddress(address = '') {
   return `${address.slice(0, 8)}...${address.slice(-4)}`
 }
+
+export function getPercentageAmount<T extends number | bigint>(amount: T, percentage = 0.99, minAmount = 0.01): T {
+  if (typeof amount === 'bigint') {
+    const percentageBigInt = BigInt(Math.floor(percentage * 10000))
+    const result = (amount * percentageBigInt) / 10000n
```
williamchong: Why `10000` not `100` here?

components/AmountInputModal.vue
nwingt: Updated

composables/use-governance-data.ts
nwingt: Updated

utils/index.ts
nwingt: Updated

utils/index.ts
```diff
@@ -67,3 +67,16 @@ export function parseURIString<T>(uri: string): T | undefined {
...
+  if (typeof amount === 'bigint') {
+    const percentageBigInt = BigInt(Math.floor(percentage * 100))
+    const result = (amount * percentageBigInt) / 100n
+    const minAmountBigInt = BigInt(Math.floor(minAmount * 100))
+    return (result > minAmountBigInt ? result : minAmountBigInt) as T
+  }
+  return Math.max(
+    Math.floor((amount as number) * percentage * 100) / 100,
+    minAmount,
+  ) as T
```
williamchong, line 81: Should both be * 10000 and / 10000 according to copilot? If we use 2dp for percentage display

pages/account/deposit.vue
```diff
@@ -300,6 +348,18 @@ const isAutoRestakeEnabled = ref(true)
 const isLoading = ref(false)
 const error = ref<string | null>(null)
 
+const maxDepositButtonTooltipText = computed(() => {
+  return Number(formatUnits(getPercentageAmount(likeBalance.value, 0.99), likeCoinTokenDecimals))
```
williamchong, line 352: Maybe `0.99` and `99%` above should be a ref/computed value

composables/use-governance-data.ts
```diff
@@ -89,18 +89,31 @@ export function useGovernanceData(walletAddress: string | Ref<string>) {
...
+  const estimatedRewardAPY = computed(() => {
+    if (!veLikeBalance.value || !estimatedRewardPerDay.value) {
+      return 0
     }
+    const dailyRate = estimatedRewardPerDay.value / Number(formatUnits(veLikeBalance.value, likeCoinTokenDecimals))
+    return Math.pow(1 + dailyRate, 365) - 1
+  })
+
+  const formattedEstimatedRewardAPY = computed(() => {
+    return (estimatedRewardAPY.value * 100).toFixed(2).concat('%')
```
williamchong, line 116: After discussion: Please keep using APY, but compound per year instead of daily

williamchong (approved):

PR #526, opened 2025-12-04
[TAS-4629] 🚸 Reset purchasing state when login fails

AuroraHuang22:


nwingt (approved):

pages/store/[nftClassId]/index.vue
```diff
@@ -1157,7 +1157,10 @@ async function handlePurchaseButtonClick() {
     isPurchasing.value = true
     if (!hasLoggedIn.value) {
       await accountStore.login()
-      if (!hasLoggedIn.value) return
+      if (!hasLoggedIn.value) {
+        isPurchasing.value = false
```
Comment (williamchong) [inline pages/store/[nftClassId]/index.vue]: should we just add a finally block to set this?

williamchong (approved):

PR #531, opened 2025-12-04
💄 Update about with staking and links

williamchong:
image


pages/about.vue
```diff
@@ -50,10 +50,12 @@
                 size="24"
                 class="text-primary"
               />
-              <h3
-                class="text-xl font-semibold text-gray-900"
-                v-text="$t('about_page_feature_ai_narration')"
-              />
+              <NuxtLink :to="localeRoute({ name: 'member' })">
+                <h3
+                  class="text-xl font-semibold text-gray-900 hover:text-primary hover:underline"
```
nwingt: actually all class should be moved to NuxtLink? especially hover:

pages/about.vue
```diff
@@ -151,10 +185,20 @@
...
+        <i18n-t
+          keypath="about_page_platform_evolution"
+          tag="p"
           class="text-lg text-gray-700 text-center leading-relaxed"
-          v-text="$t('about_page_platform_evolution')"
-        />
+        >
+          <template #likecoin_v3>
+            <a
+              href="https://dao.like.co"
```
nwingt: should use like.co?

nwingt (approved):

PR #544, opened 2025-12-10
[TAS-5047] 👔 Do not show $1 for 30 days in yearly plan for upsell

nwingt:
Screenshot 2025-12-11 at 01 16 04
Screenshot 2025-12-11 at 01 43 11


components/UpsellPlusModal.vue
```diff
@@ -95,6 +95,7 @@
         class="w-full mt-4 laptop:mt-6"
         :is-monthly-hidden="!shouldShowMonthlyPlan"
         :is-yearly-hidden="!shouldShowYearlyPlan"
+        :is-upsell="true"
```
williamchong: if gift nftClassId is not set we still want to show $1/30d for yearly, can we expose more control here to do that?

williamchong (changes requested):

components/UpsellPlusModal.vue
nwingt: updated

williamchong (approved):

PR #551, opened 2025-12-12
[TAS-5056] 👔 Don't show monthly upsell if from channel is set

williamchong:


pages/store/[nftClassId]/index.vue
```diff
@@ -1216,6 +1216,7 @@ async function handlePurchaseButtonClick() {
         utmSource: 'upsell_plus',
         utmCampaign: `upsell_plus_${nftClassId.value}`,
         utmMedium: 'product_page',
+        from: getRouteQuery('from') as string | undefined,
```
Comment (nwingt) [inline pages/store/[nftClassId]/index.vue]: do you mean this? as `getRouteQuery()` always returns a string
```suggestion
        from: getRouteQuery('from') || undefined,
```

nwingt (approved):

PR #563, opened 2025-12-17
[TAS-5083] 🐛 Check period match when refreshing plus status

williamchong:


composables/use-subscription.ts
```diff
@@ -174,7 +174,10 @@ export function useSubscription() {
           period: subscribePlan,
           giftNFTClassId: isYearly ? nftClassId : undefined,
         })
-        await navigateTo(localeRoute({ name: 'plus-success' }))
+        if (upsellPlusModal.isOpen) {
+          upsellPlusModal.close()
+        }
```
nwingt: is it more robust to watch route change and close the modal?

nwingt (approved):

composables/use-subscription.ts
williamchong: seems it still doesn’t close at all in this commit though

PR #568, opened 2025-12-18
✨ Store read progress and sort by last open in shelf

williamchong:
image

### DIFF
--- a/components/BookshelfItem.vue
+++ b/components/BookshelfItem.vue
@@ -14,7 +14,24 @@
       @click="handleCoverClick"
     />
 
-    <div class="mt-2 h-[70px]">
+    <div class="mt-2 w-full h-[24px]">
+      <div
+        v-if="progressPercentage > 0 && progressPercentage <= 100"
+        class="w-full"
+      >
+        <div class="flex items-center justify-between text-xs text-toned mb-1">
+          <span>{{ progressPercentage }}%</span>
+        </div>
+        <div class="w-full h-1 bg-gray-200 rounded-full overflow-hidden">
+          <div
+            class="h-full bg-primary-500 transition-all duration-300"
+            :style="{ width: `${progressPercentage}%` }"
+          />
+        </div>
+      </div>
+    </div>
+
+    <div class="h-[70px]">
       <div class="flex items-start gap-1">
         <div
           class="text-sm laptop:text-base text-highlighted font-semibold line-clamp-2 grow"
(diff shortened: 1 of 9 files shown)


composables/use-reader-progress.ts
```diff
@@ -0,0 +1,39 @@
...
+    0,
+  )
+
+  onMounted(() => {
+    lastOpenedTime.value = Date.now()
+  })
+
+  onBeforeUnmount(() => {
+    bookshelfStore.updateProgress(nftClassId, readProgress.value, lastOpenedTime.value)
+  })
```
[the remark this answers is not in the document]
williamchong: Back button in reader doesn't make shelf reload from storage

components/BookshelfItem.vue
```diff
@@ -14,7 +14,24 @@
...
 
-    <div class="mt-2 h-[70px]">
+    <div class="mt-2 w-full h-[24px]">
+      <div
+        v-if="progressPercentage > 0 && progressPercentage <= 100"
+        class="w-full"
+      >
+        <div class="flex items-center justify-between text-xs text-toned mb-1">
+          <span>{{ progressPercentage }}%</span>
+        </div>
```
nwingt: since only one child, no need to add flex related CSS

also decrease margin bottom
```suggestion
        <div
          class="text-xs text-toned mb-0.5"
          v-text="`${progressPercentage}%`"
        />
```

components/BookshelfItem.vue
```diff
@@ -14,7 +14,24 @@
       @click="handleCoverClick"
     />
 
-    <div class="mt-2 h-[70px]">
+    <div class="mt-2 w-full h-[24px]">
```
nwingt: I think add little spacing is better, it is too tight now



```suggestion
    <div class="mt-2 mb-1 w-full h-[24px]">
```

composables/use-reader-progress.ts
```diff
@@ -0,0 +1,39 @@
...
+  nftClassId: string
+  bookProgressKeyPrefix: string
+}) {
+  const bookshelfStore = useBookshelfStore()
+
+  function getProgressKeyWithSuffix(suffix: ReaderCacheKeySuffix) {
+    return getReaderCacheKeyWithSuffix(bookProgressKeyPrefix, suffix)
+  }
+
+  const readProgress = useStorage(
```
nwingt:
```suggestion
  const readingProgress = useStorage(
```

PR #589, opened 2026-01-02
♻️ Refactor trial days config

williamchong:


constants/pricing.ts
```diff
@@ -207,3 +207,7 @@ export const TWD_PRICE_TIER_LIST = [
   2970,
   3000,
 ]
+
+export const DEFAULT_TRIAL_PERIOD_DAYS = 7
+export const PAID_TRIAL_PERIOD_DAYS = 30
+export const DEFAULT_TRIAL_PRICE = 1
```
nwingt: `DEFAULT_TRIAL_PRICE` should be 0 now?

i am not sure whether we should create a map like with days as key and price as value

williamchong: Rewrote the concept to `paid trial thersold` according to current upsell ui logic

nwingt (approved):

PR #591, opened 2026-01-05
✨ Add access info in user collection doc

williamchong:


app.vue
```diff
@@ -21,6 +23,28 @@ const isTestnet = !!config.public.isTestnet
 const { memberProgramData } = useMemberProgramStructuredData()
 
 const i18nHead = useLocaleHead()
+
+const { user } = useUserSession()
+const accountStore = useAccountStore()
+const lastSessionRefresh = useStorage('lastSessionRefresh', 0)
```
nwingt:
```suggestion
const lastSessionRefreshTs = useStorage('lastSessionRefreshTs', 0)
```

nwingt (approved):

PR #590, opened 2026-01-05
♻️ Refactor country detection

williamchong:

PR #592, opened 2026-01-05
🚸 Use cf-ipcountry for locale detection

williamchong:
Base on https://github.com/likecoin/3ook-com/pull/590


nwingt:
i am not sure whether want to have locale detection for 3ook.com

cc @ckxpress

nwingt (approved):

PR #596, opened 2026-01-07
🔍️ Add 電子書 in title

williamchong:


nwingt (approved):

nwingt:
so only book with `電子書` keyword will show 電子書 in title, right?
but I guess most of the books dont have `電子書` in keywords?

williamchong:
> keywords

yes should use `descriptionTags` instead..

PR #597, opened 2026-01-07
[TAS-4441] 💩 Add burn NFT page

williamchong:
image


pages/account/burn.vue
```diff
@@ -0,0 +1,302 @@
...
+      <UCard
+        v-else
+        class="w-full max-w-sm mx-auto mt-8"
+        :ui="{ footer: 'flex justify-end' }"
+      >
+        <p v-text="$t('burn_page_please_login')" />
+        <template #footer>
+          <LoginButton />
+        </template>
+      </UCard>
```
nwingt: there is a login button in `/account.vue`, I guess we can remove the login button from the child page

pages/account/burn.vue
```diff
@@ -0,0 +1,302 @@
...
+          </div>
+        </UCard>
+
+        <UAlert
+          icon="i-material-symbols-warning-outline-rounded"
+          color="error"
+          variant="subtle"
+          :title="$t('burn_page_warning_title')"
+          :description="$t('burn_page_warning_description')"
+        />
```
nwingt: I think the alert should be placed at the top
cant see it if there is a long list of NFTs

pages/account/burn.vue
```diff
@@ -0,0 +1,302 @@
...
+        :ui="{ footer: 'flex justify-end' }"
+      >
+        <p v-text="$t('burn_page_please_login')" />
+        <template #footer>
+          <LoginButton />
+        </template>
+      </UCard>
+    </main>
+
+    <UModal
```
nwingt, line 94: it seems this modal can be placed within the `<main/>` and then the root `<div/>` can be removed
wanna use `<main/>` as the root element for all child pages under `/account`, similar to `/account/deposit`

PR #605, opened 2026-01-10
💄 Add author video in about page

williamchong:
image


pages/about.vue
```diff
@@ -211,6 +211,98 @@
...
+          <h2 class="text-2xl md:text-3xl font-bold text-gray-900">
+            {{ $t('about_page_author_stories_title') }}
+          </h2>
+          <p class="text-lg text-gray-600 leading-relaxed max-w-2xl mx-auto">
+            {{ $t('about_page_author_stories_content') }}
+          </p>
+        </div>
+
+        <UCard class="overflow-hidden">
+          <div class="space-y-4">
```
nwingt: I think this class can be put in `<UCard/>` body then remove this redundant div

pages/about.vue
```diff
@@ -211,6 +211,98 @@
...
+              <div>
+                <h3 class="text-xl font-bold text-gray-900">
+                  <NuxtLink
+                    :to="localeRoute({ name: 'store', query: { author: '董啟章' } })"
+                    class="hover:text-primary hover:underline"
+                  >
+                    {{ $t('about_page_author_dong_qizhang') }}
+                  </NuxtLink>
+                </h3>
+              </div>
```
nwingt: what is the purpose of wrapping h3 with div?

pages/about.vue
```diff
@@ -211,6 +211,98 @@
...
+        <UCard class="overflow-hidden">
+          <div class="space-y-4">
+            <div class="flex items-center gap-3">
+              <div class="w-12 h-12 rounded-full bg-primary/10 flex items-center justify-center">
+                <UIcon
+                  name="i-material-symbols-person-rounded"
+                  size="24"
+                  class="text-primary"
+                />
+              </div>
```
nwingt: how about using `<UAvatar/>` with icon prop? we can add profile pic later

```suggestion
              <UAvatar
                icon="i-material-symbols-person-rounded"
                size="md"
              />
```

pages/about.vue
```diff
@@ -211,6 +211,98 @@
...
+              <div>
+                <h3 class="text-xl font-bold text-gray-900">
+                  <NuxtLink
+                    :to="localeRoute({ name: 'store', query: { author: '董啟章' } })"
+                    class="hover:text-primary hover:underline"
+                  >
+                    {{ $t('about_page_author_dong_qizhang') }}
+                  </NuxtLink>
```
nwingt: for all kinds of links, it is better using `<UButton/>` with variant `link`. Then, you dont have to style it and also it has paddings which increase its click zone on mobile

i18n/locales/en.json
```diff
@@ -523,4 +529,4 @@
   "upsell_plus_yearly_member": "Yearly",
   "upsell_plus_yearly_notice": "Become a Plus {year} member and you will {gift}",
   "work_in_progress": "Work in progress"
-}
+}
```
nwingt: why removing the new line at the end? do you not using i18n-ally?

nwingt (changes requested):

i18n/locales/en.json
williamchong: vscode sort json function somehow always remove newline...

nwingt (approved):

pages/about.vue
```diff
@@ -211,6 +211,91 @@
...
+                variant="link"
+                :label="$t('about_page_author_dung_kai_cheung')"
+              />
+            </h3>
+          </div>
+          <p class="text-gray-600 leading-relaxed">
+            {{ $t('about_page_author_dung_kai_cheung_desc') }}
+          </p>
+
+          <div class="aspect-video w-full rounded-xl overflow-hidden bg-gray-900 shadow-inner -mx-4 md:mx-0">
```
nwingt: offset in mobile


why adding negative margin?
```suggestion
          <div class="aspect-video w-full rounded-xl overflow-hidden bg-gray-900 shadow-inner">
```

nwingt (changes requested):

PR #608, opened 2026-01-12
[TAS-5174] 👽 Prepare for native app submission

nwingt:


pages/list.vue
```diff
@@ -20,16 +20,27 @@
...
               @update:model-value="handleSelectAllUpdate"
             />
           </div>
-          <div class="flex items-center justify-between col-start-2 col-span-11">
+          <div
+            :class="[
+              'flex',
+              'items-center',
+              'justify-between',
+              { 'col-span-11': !isApp },
```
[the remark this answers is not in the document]
nwingt: intended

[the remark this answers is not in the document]
nwingt: intended

pages/account/index.vue
```diff
@@ -244,6 +251,7 @@
         </AccountSettingsItem>
 
         <AccountSettingsItem
+          v-if="!isApp || user?.isLikerPlus"
```
williamchong: Can we still show a disabled color mode button and says plus required, without any action on click?

pages/account/index.vue
```diff
@@ -295,6 +303,7 @@
         />
 
         <UButton
+          v-if="!isApp"
```
williamchong, line 308: Not sure if we can/should show a /member page that has feature list, but no pricing info

composables/use-app-detection.ts
```diff
@@ -0,0 +1,15 @@
+export function useAppDetection() {
+  const getRouteQuery = useRouteQuery()
+
+  const isApp = computed(() => {
+    let isAppUserAgent = false
+    if (import.meta.client) {
+      isAppUserAgent = navigator.userAgent?.startsWith('3ook-com-app') || false
+    }
```
williamchong: Not sure if this will cause SSR hydration mismatch issue, maybe check userAgent onMounted instead

components/TTSPlayerModal.vue
```diff
@@ -277,6 +278,13 @@ const {
       && error.code === MediaError.MEDIA_ERR_SRC_NOT_SUPPORTED
       && !user.value?.isLikerPlus) {
       stopTextToSpeech()
+      if (isApp.value) {
+        errorModal.open({
+          title: $t('tts_free_trial_limit_error_title'),
+          description: $t('tts_free_trial_limit_error_description'),
+        })
+        return
+      }
```
williamchong, line 287: what about we show paywall modal without any pricing info and cta

williamchong (changes requested):

williamchong (approved):

PR #609, opened 2026-01-12
[TAS-5202] ✨ Sync user setting to db

williamchong:


composables/use-auto-locale.ts
```diff
@@ -18,29 +18,39 @@ function getDefaultLocaleFromCountry(country: string | null): LocaleCode {
...
+  const detectedLocale = computed(() => getDefaultLocaleFromCountry(detectedCountry.value))
 
   function setLocale(locale: LocaleCode) {
-    storedLocale.value = locale
+    if (hasLoggedIn.value) {
+      syncedLocale.value = locale
+    }
+    else {
+      localStorageLocale.value = locale
+    }
```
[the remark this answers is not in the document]
nwingt: @williamchong make sense

it is strange that currency has different logic

composables/use-synced-user-settings.ts
```diff
@@ -0,0 +1,45 @@
...
+  watch(hasLoggedIn, (isLoggedIn) => {
+    if (isLoggedIn && !userSettingsStore.isInitialized()) {
+      userSettingsStore.ensureInitialized()
+    }
+  })
+
+  onMounted(() => {
+    if (hasLoggedIn.value && !userSettingsStore.isInitialized()) {
+      userSettingsStore.ensureInitialized()
+    }
```
nwingt: Maybe refactor these lines to share with the watcher above?

server/api/user/settings.post.ts
```diff
@@ -0,0 +1,28 @@
+import type { UserSettingKey } from '~/types/user-settings'
```
nwingt: maybe move the user-settings to `/shared/types` as both client and server side using it

shared/types/user-settings.d.ts
```diff
@@ -0,0 +1,7 @@
+import type { LocaleCode } from '~/composables/use-auto-locale'
+import type { PaymentCurrency } from '~/composables/use-payment-currency'
```
nwingt: how about move the type to `/shared/types`

PR #610, opened 2026-01-12
💫 Add animation when claiming book

williamchong:
image


nwingt:
suppose the white glare not overflowing the cover?

williamchong:
> suppose the white glare not overflowing the cover? Screenshot 2026-01-12 at 23 02 13 Screenshot 2026-01-12 at 23 01 12

Ideally they shouldn't, but seems I can't line up them properly in `BookLoadingScreen`, alsp seems not suitable to go into `BookCover`

PR #624, opened 2026-01-16
💫 Add animation for reader loading

williamchong:
I also want to remove all text on loaded, but somehow doesn't work, lets add a dumb transition first

### DIFF
--- a/pages/reader/epub.vue
+++ b/pages/reader/epub.vue
@@ -1,13 +1,20 @@
 <template>
   <main>
-    <BookLoadingScreen
-      v-if="isReaderLoading"
-      class="absolute inset-0"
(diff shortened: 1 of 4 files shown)


pages/reader/epub.vue
```diff
@@ -1,13 +1,22 @@
...
-      v-if="isReaderLoading"
-      class="absolute inset-0"
-      :book-name="bookInfo.name.value"
-      :book-cover-src="bookCoverSrc"
-      :loading-label="loadingLabel"
-    />
-    <div class="relative flex flex-col w-full grow">
+    <Transition
+      name="reader-load"
+    >
```
nwingt:
```suggestion
    <Transition name="reader-load">
```

PR #626, opened 2026-01-19
🐛 Fix invalid check in unstake

williamchong:


components/StakingControl.vue
```diff
@@ -283,7 +283,7 @@ async function handleUnstakeButtonClick() {
     helpText: $t('amount_available', { amount: `${formattedUserStake.value} ${likeCoinTokenSymbol}` }),
   }).result
 
-  if (!isValidStakeAmount.value) return
+  if (stakeAmount.value <= 0) return
```
nwingt, line 286: the same in `handleDonateButtonClick()` below

williamchong: donate is valid, because it spend current LIKE balance
unstake is invalid, because it convert staked LIKE amount to LIKE balance

nwingt: isnt donate is donating the rewards?

williamchong: donate is donating current LIKE balance, to increase the reward pool

nwingt: very misun

nwingt (approved):

components/StakingControl.vue
williamchong: donate_to_reward maybe 
the button was made for debugging only anyway..

PR #630, opened 2026-01-21
💄 Fix PDF reader flashing when transiting in

williamchong:


nwingt (changes requested):
`justify-center` just doesn't work

https://github.com/user-attachments/assets/ba0f18d4-e391-4561-9d39-ac003f177a9a

williamchong:
> `justify-center` just doesn't work
> 
>  trim.1B352067-B11D-43FE-88ED-DEADE2BD8CF9.MOV

lets just do " Fix PDF reader flashing when transiting in" then

pages/reader/pdf.vue
```diff
@@ -1,15 +1,19 @@
...
-      />
+        class="absolute inset-0 z-10 bg-background"
+      >
+        <BookLoadingScreen
+          :book-name="bookInfo.name.value"
+          :book-cover-src="bookCoverSrc"
+          :loading-label="loadingLabel"
+          :loading-progress="loadingPercentage"
+        />
+      </div>
```
nwingt: should be the same?
```suggestion
      <BookLoadingScreen
        v-if="isReaderLoading"
        class="absolute inset-0 z-10 bg-background"
        :book-name="bookInfo.name.value"
        :book-cover-src="bookCoverSrc"
        :loading-label="loadingLabel"
        :loading-progress="loadingPercentage"
      />
```

nwingt (approved):

PR #636, opened 2026-01-22
🚸 Auto pause TTS on offline

williamchong:
閱讀 《佛繫人生》 - 何昊禧  3ook com


components/TTSPlayerModal.vue
```diff
@@ -179,6 +179,42 @@
...
+              class="flex-1"
+              @click="handleOfflineModalForceResume"
+            >
+              {{ $t('tts_offline_modal_resume_button') }}
+            </UButton>
+          </div>
+        </div>
+      </div>
+    </template>
+  </UModal>
```
nwingt: can you teach Claude code to use header, body and footer slot of UModal?

williamchong: It is copying the model directly above, which doesn't work so I merged them now
Should we maintain and commit an `AGENTS.md` in repo, like eslint/tsconfig rules

nwingt: I think we should

i18n/locales/en.json
```diff
@@ -522,7 +522,11 @@
   "tts_sample_mandarin_description": "Read by Lei Ting Yin",
   "tts_samples_section_footer": "Subscribe to Plus and unlock more voices to enjoy.",
   "tts_samples_section_title": "Try AI Text-to-Speech",
-  "tts_try_modal_description": "On 3ook.com, books aren’t just for reading — you can listen to them too.\nLet our voice actors read them for you.",
+  "tts_offline_modal_description": "TTS has been paused because the next track could not be loaded without internet. Playback will resume automatically when your connection is restored.",
+  "tts_offline_modal_resume_button": "Try Anyway",
+  "tts_offline_modal_stop_button": "Stop Playing",
+  "tts_offline_modal_title": "No Internet Connection",
```
nwingt: Please sort keys

components/TTSPlayerModal.vue
```diff
@@ -1,11 +1,46 @@
...
+              {{ $t('tts_offline_modal_resume_button') }}
+            </UButton>
+          </div>
+        </div>
+      </div>
+
+      <div
+        v-else
+        class="flex flex-col h-full w-full max-w-[670px] mx-auto py-6 px-8 laptop:px-4"
+      >
```
nwingt, line 20: Perhaps we should put v-if on UModal level and not mess with this TTS custom modal with content slot

nwingt: maybe just use ErrorModal for offline modal

nwingt (approved):

PR #661, opened 2026-02-04
🚸 Link plus features to related links

williamchong:
image

### DIFF
--- a/components/PricingPlanBenefits.vue
+++ b/components/PricingPlanBenefits.vue
@@ -36,16 +36,35 @@
       </li>
       <li>
         <UIcon name="i-material-symbols-check" />
-        <span v-text="$t('pricing_page_feature_2')" />
+        <NuxtLink
+          :to="localeRoute({ name: 'store' })"
+          target="_blank"
+          class="underline"
+          @click="useLogEvent('pricing_benefit_click_store')"
+        >
+          {{ $t('pricing_page_feature_2') }}
+        </NuxtLink>
       </li>
       <template v-if="!isCompact">
         <li>
           <UIcon name="i-material-symbols-check" />
-          <span v-text="$t('pricing_page_feature_3')" />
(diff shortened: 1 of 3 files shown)


components/PricingPlanBenefits.vue
```diff
@@ -36,16 +36,32 @@
       </li>
       <li>
         <UIcon name="i-material-symbols-check" />
-        <span v-text="$t('pricing_page_feature_2')" />
+        <NuxtLink
+          :to="localeRoute({ name: 'store', query: { ll_medium: 'plus-benefit', ll_source: 'plus-modal' } })"
+          target="_blank"
```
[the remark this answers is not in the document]
nwingt: this component is used in other places, I think it is better to set `ll_source` from props or just leave it

Image

i18n/locales/en.json
```diff
@@ -370,6 +370,7 @@
   "pricing_page_feature_2": "20% off all book purchases, no limit",
   "pricing_page_feature_3": "Early access to new features",
   "pricing_page_feature_4": "Live reply to customer service messages",
+  "pricing_page_intercom_prefill": "Hi, I'd like to learn more about 3ook Plus",
```
nwingt: please sort keys

i18n/locales/zh-Hant.json
```diff
@@ -370,6 +370,7 @@
   "pricing_page_feature_2": "購書八折，不設上限",
   "pricing_page_feature_3": "優先使用新功能",
   "pricing_page_feature_4": "真人回覆客服訊息",
+  "pricing_page_intercom_prefill": "你好，我想了解更多關於 3ook Plus 的資訊",
```
nwingt: please sort keys

nwingt (changes requested):

nwingt (approved):

PR #680, opened 2026-02-09
💄 Simplify about page CTA

williamchong:
image


nwingt:
試聽 button seems not enough contrast from your screenshot

williamchong:
> 試聽 button seems not enough contrast from your screenshot

updated, not sure what class to use to make it dark mode friendly though

pages/about.vue
```diff
@@ -36,17 +36,67 @@
...
+          color="primary"
+          size="xl"
+          icon="i-material-symbols-store-rounded"
+          @click="onClickHeroCtaStore"
+        />
+        <UButton
+          :to="localeRoute({ name: 'member', query: { samples: '1' } })"
+          :label="$t('about_page_hero_cta_narration')"
+          variant="outline"
+          class="text-white ring-white hover:bg-white/10"
```
nwingt: i think it is better to use text-inverted and ring-inverted

nwingt (approved):

PR #681, opened 2026-02-09
🚸 Add about links in blocking pages

williamchong:
image
image
image


pages/shelf/[[walletAddress]].vue
```diff
@@ -38,12 +38,30 @@
...
+        />
+        <LoginButton class="mt-5" />
+        <UButton
+          class="mt-2"
+          :label="$t('bookshelf_please_login_learn_more')"
+          :to="localeRoute({ name: 'about', query: { ll_medium: 'about-link', ll_source: 'shelf' } })"
+          variant="link"
+          color="neutral"
+          size="sm"
+        />
```
Comment (nwingt) [inline pages/shelf/[[walletAddress]].vue]: how about putting these two buttons in the footer slot, [about] [login], with flex justify-end class

i18n/locales/zh-Hant.json
```diff
@@ -499,7 +502,10 @@
   "store_fetch_more_items_error": "無法載入更多書本",
   "store_fetch_tags_error": "無法載入分類",
   "store_no_items": "沒有書本",
+  "store_no_items_learn_more": "了解更多關於 3ook.com",
   "store_no_search_results": "沒有找到結果",
+  "store_no_search_results_contact": "聯絡我們，幫你找到想要的書",
```
nwingt: i think it is too long for a button label, not sure this has enough space in mobile screen

how about：
需要我們幫你找想要的書？
[聯絡我們]

pages/shelf/[[walletAddress]].vue
Comment (williamchong) [reply pages/shelf/[[walletAddress]].vue]: image
like this?

nwingt (approved):

PR #684, opened 2026-02-10
💄 Simplify default paywall modal

williamchong:
image

### DIFF
--- a/components/PaywallModal.vue
+++ b/components/PaywallModal.vue
@@ -122,6 +122,7 @@
             'max-laptop:mx-auto',
             'p-5 laptop:p-12',
             'pt-12',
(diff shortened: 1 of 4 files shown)


components/PaywallModal.vue
```diff
@@ -122,6 +122,7 @@
             'max-laptop:mx-auto',
             'p-5 laptop:p-12',
             'pt-12',
+            { 'laptop:pt-16': !campaignContent },
```
nwingt, line 126: Can we add a note here?
```suggestion
            // NOTE: Prevent content from being covered by the Intercom banner at the top
            { 'laptop:pt-16': !campaignContent },
```

nwingt (approved):

PR #696, opened 2026-02-12
[TAS-5324] ✨ Add account delete button

nwingt:
Depends on https://github.com/likecoin/likecoin-api-public/pull/1195


williamchong (approved):
This likely won't pass apple review, need to be actual automated API

PR #701, opened 2026-02-13
🔒 Switch locale sync to /preferences API with scoped permissions

williamchong:
The /users/update endpoint requires 'write' scope which our token doesn't have. Switch to /users/preferences which only needs 'write:preferences', and add read/write:preferences scopes to login JWT permissions.

### DIFF
--- a/server/api/user/settings.post.ts
+++ b/server/api/user/settings.post.ts
@@ -1,3 +1,5 @@
+import { jwtDecode } from 'jwt-decode'
+
 const ALLOWED_KEYS = ['locale', 'currency', 'colorMode'] as const satisfies readonly UserSettingKey[]
 
 export default defineEventHandler(async (event) => {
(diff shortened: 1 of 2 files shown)


server/api/user/settings.post.ts
```diff
@@ -38,18 +40,22 @@ export default defineEventHandler(async (event) => {
...
+        await getLikeCoinAPIFetch()('/users/preferences', {
+          method: 'POST',
+          headers: {
+            Authorization: `Bearer ${session.user.token}`,
+          },
+          body: {
+            locale: body.locale?.startsWith('zh') ? 'zh' : 'en',
+          },
+        })
+      }
```
nwingt, line 58: should we need to handle else case?

nwingt (approved):

server/api/user/settings.post.ts
williamchong: just let it be for now, best effort update

PR #716, opened 2026-02-20
♻️ Refactor to use VueUse composables over manual implementations

williamchong:
Replace manual patterns with VueUse equivalents for better maintainability:
- useLocalStorage for localStorage in account store (plusRedirectRoute)
- useResizeObserver in ExpandableContent (removes manual lifecycle)
- useEventListener in PDFReader (removes addEventListener/removeEventListener)
- useVModel in PaywallModal (simplifies v-model wiring)
- useIntervalFn in CustomVoiceUploadModal (replaces setInterval/clearInterval)
- watchImmediate in AffiliateAlert and account page (cleaner API)

### DIFF
--- a/composables/use-subscription.ts
+++ b/composables/use-subscription.ts
@@ -199,7 +199,7 @@ export function useSubscription() {
           coupon,
         })
         if (redirectRoute && redirectRoute?.name) {
-          accountStore.plusRedirectRoute = redirectRoute
+          accountStore.savePlusRedirectRoute(redirectRoute)
         }
         await navigateTo(url, { external: true })
       }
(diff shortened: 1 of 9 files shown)


composables/use-subscription.ts
```diff
@@ -199,7 +199,7 @@ export function useSubscription() {
           coupon,
         })
         if (redirectRoute && redirectRoute?.name) {
-          accountStore.savePlusRedirectRoute(redirectRoute)
+          accountStore.plusRedirectRoute = redirectRoute
```
nwingt: not sure whether this is a good practice to alter state of store without using action

nwingt (approved):

PR #725, opened 2026-02-24
✨ Add prompt audio to custom voice cloning

williamchong:
Support short phrase (prompt voice) upload alongside main voice sample when cloning voices via MiniMax API, improving voice similarity.


components/CustomVoiceUploadModal.vue
```diff
@@ -334,170 +431,143 @@ const previewAvatarUrl = computed(() => {
...
+const promptAudioInputEl = useTemplateRef<HTMLInputElement>('promptAudioInputEl')
+
+const voiceLanguageOptions = computed(() =>
+  locale.value === 'en'
+    ? [{ label: 'English', value: 'en-US' }]
+    : [
+        { label: '粵語', value: 'zh-HK' },
+        { label: '國語', value: 'zh-TW' },
+      ],
+)
```
nwingt, line 515: should we show all options regardless of the current locale?

components/CustomVoiceUploadModal.vue
```diff
@@ -80,57 +83,93 @@
...
-          </blockquote>
         </div>
 
+        <!-- Step 1: Main voice sample -->
         <div class="flex flex-col gap-2">
-          <label
-            class="text-sm font-medium"
-            v-text="$t('tts_custom_voice_audio_source_label')"
+          <label class="text-sm font-medium">
+            1. {{ $t('tts_custom_voice_audio_source_label') }}
```
nwingt, line 130: should we include 1. in locale string too?

nwingt (approved):

components/CustomVoiceUploadModal.vue
williamchong: We show english text for user to read, and chinese for zh, so showing all is confusing

PR #732, opened 2026-02-25
🐛 Dont init farcaster in 3ook App

williamchong:


nwingt (approved):

ckxpress:
curious why we init'ed farcaster previously? coinbase lib?

williamchong:
> curious why we init'ed farcaster previously? coinbase lib?

Want to be miniapp in base but doesn't rly work meaningfully

PR #733, opened 2026-02-25
[TAS-5358] ✨ Support isAdultOnly: hide 18+ books by default, add opt-in toggle o…

williamchong:
…n account page

- Add isAdultOnly to BookstoreInfo type and isAdultContentEnabled to user settings
- Filter adult-only content from store search results, staking listings, and recommendations
- Show 18+ badge on product detail page when book is marked adult-only
- Add account page toggle with age verification confirmation modal
- Add i18n labels for zh-Hant and en


pages/account/index.vue
```diff
@@ -425,6 +442,39 @@
...
+    <UModal
+      v-model:open="isAdultContentConfirmOpen"
+      :ui="{
+        title: 'text-lg font-bold',
+        footer: 'flex justify-end gap-3',
+      }"
+    >
+      <template #title>
+        <span v-text="$t('account_page_adult_content_confirm_title')" />
+      </template>
```
nwingt: simply use title prop?

pages/account/index.vue
```diff
@@ -425,6 +442,39 @@
...
+
+      <template #body>
+        <p
+          class="text-sm"
+          v-text="$t('account_page_adult_content_confirm_description')"
+        />
+      </template>
```
nwingt: simply use description prop?

nwingt (approved):

PR #738, opened 2026-02-26
🐛 Fix iOS background play and media session

williamchong:


composables/use-web-audio-player.ts
```diff
@@ -57,15 +60,10 @@ export function useWebAudioPlayer(): TTSAudioPlayer {
...
         pausedInternally = false
-        handlers.pause?.()
         return
       }
       handlers.pause?.()
-      // Don't auto-resume if the audio element is in an error state;
-      // the error handler will skip to the next segment instead.
       if (errored) return
-      // Unexpected pause (OS interruption: phone call, other app audio, etc.)
-      // Attempt auto-resume after a short delay, with a retry limit
```
nwingt, line 68: why remove comment if nothing changes?

nwingt (approved):

PR #746, opened 2026-02-27
♻️ Extract PricingPageContent from PaywallModal for inline member pag…

williamchong:
…e rendering

The member page previously opened PaywallModal as a fullscreen overlay hack. Now it renders PricingPageContent directly as page content, while PaywallModal remains a thin modal wrapper for other callers (e.g. TTS player upsell).

### DIFF
--- a/pages/member.vue
+++ b/pages/member.vue
@@ -1,5 +1,28 @@
 <template>
-  <AppHeader :is-connect-hidden="false" />
+  <PricingPageContent
+    v-model="selectedPlan"
+    class="min-h-screen"
+    :is-processing-subscription="subscription.isProcessingSubscription.value"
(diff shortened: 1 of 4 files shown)


pages/member.vue
```diff
@@ -1,5 +1,30 @@
...
+          class="absolute z-10 top-2 phone:top-4 left-2 phone:left-4 text-white cursor-pointer"
+          icon="i-material-symbols-arrow-back"
+          :to="localeRoute({ name: 'store' })"
+          variant="ghost"
+          color="neutral"
+          size="md"
+        />
+      </template>
+    </PricingPageContent>
+  </div>
```
nwingt: Can remove extra div?

nwingt (approved):

PR #749, opened 2026-02-27
👔 Hide subscription entry in about page for app

nwingt:


williamchong (changes requested):
also need to hide the button near onClickHeroCtaNarration

williamchong (approved):

PR #754, opened 2026-03-01
📈 Add server-side PubSub event publishing for Elasticsearch analytics

williamchong:
Publish fire-and-forget events to GCP PubSub `misc` topic with `3ook_` prefix, mirroring the pattern from liker-land v1. Events cover auth, book lists, TTS, annotations, user settings, and custom voice actions. Guarded by GCLOUD_PUBSUB_ENABLE env var — silently skipped when unset.

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -54,6 +54,8 @@ export default defineNuxtConfig({
   },
 
   runtimeConfig: {
+    pubsubEnable: process.env.GCLOUD_PUBSUB_ENABLE || '',
+    appServer: process.env.APP_SERVER || '3ook-web',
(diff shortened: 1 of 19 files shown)


nuxt.config.ts
```diff
@@ -54,6 +54,8 @@ export default defineNuxtConfig({
   },
 
   runtimeConfig: {
+    pubsubEnable: process.env.GCLOUD_PUBSUB_ENABLE || '',
```
nwingt, line 57:
```suggestion
    isPubSubEnabled: process.env.IS_PUBSUB_ENABLED || '',
```

PR #755, opened 2026-03-01
🐛 Fix TTS player blocking account page when navigating to custom voic…

williamchong:
…e setup

Close TTS player modal before navigating to account page and auto-open custom voice upload modal via query parameter.

### DIFF
--- a/pages/account/index.vue
+++ b/pages/account/index.vue
@@ -557,9 +557,28 @@ function confirmAdultContent() {
 const { customVoice, hasCustomVoice, fetchCustomVoice } = useCustomVoice()
 const overlay = useOverlay()
 const customVoiceModal = overlay.create(CustomVoiceUploadModal)
+const blockingModal = useBlockingModal()
 
-watchImmediate(hasLoggedIn, (loggedIn) => {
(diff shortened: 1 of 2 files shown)


pages/account/index.vue
```diff
@@ -558,8 +558,15 @@ const { customVoice, hasCustomVoice, fetchCustomVoice } = useCustomVoice()
 const overlay = useOverlay()
 const customVoiceModal = overlay.create(CustomVoiceUploadModal)
 
-watchImmediate(hasLoggedIn, (loggedIn) => {
-  if (loggedIn) fetchCustomVoice()
+const route = useRoute()
+
+watchImmediate(hasLoggedIn, async (loggedIn) => {
+  if (loggedIn) {
+    await fetchCustomVoice()
```
nwingt: should we show blockingModal when fetching custom voices?

nwingt (approved):

pages/account/index.vue
williamchong: maybe only when `route.query.action === 'custom-voice'`

PR #758, opened 2026-03-02
🐛 Fix audio stall recovery and stuck detection for web player

williamchong:
Add stuck-detection timeout to recover when onplay never fires, and stall retry for Safari iOS getting stuck at >1x playback speed. Mark errored state on NotAllowedError to prevent duplicate error events and spurious auto-resume attempts.

### DIFF
--- a/composables/use-web-audio-player.ts
+++ b/composables/use-web-audio-player.ts
@@ -186,21 +186,17 @@ export function useWebAudioPlayer(): TTSAudioPlayer {
 
     // Stuck detection: if onplay never fires within timeout, retry once then error
     stuckTimer = setTimeout(() => {
-      if (!playing && active && !errored) {
-        if (!stuckRetried) {
-          console.warn(`Audio stuck — retrying playback`)
-          stuckRetried = true
-          audio.load()
-          audio.play()?.catch(e => handlePlayError(e, { clearStuck: true }))
-          stuckTimer = setTimeout(() => {
-            if (!playing && active && !errored) {
-              console.warn(`Audio stuck — retry failed after ${STUCK_DETECTION_TIMEOUT_MS}ms`)
-              errored = true
-              handlers.error?.(audio.error || 'STUCK_TIMEOUT')
-            }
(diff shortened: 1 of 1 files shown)


composables/use-web-audio-player.ts
```diff
@@ -99,6 +103,9 @@ export function useWebAudioPlayer(): TTSAudioPlayer {
       if (audio.currentTime < 0.00001) {
         // Safari on iOS sometimes gets stuck at 0.000001 for rate > 1.0
         audio.playbackRate = 1.0
+        if (active && !errored) {
+          audio.play()?.catch(() => {})
```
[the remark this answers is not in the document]
williamchong: same audio on >1.0 rate would get stuck again, not?

composables/use-web-audio-player.ts
```diff
@@ -154,14 +182,26 @@ export function useWebAudioPlayer(): TTSAudioPlayer {
...
-      console.warn('Play rejected:', e)
-      if (e instanceof DOMException && e.name === 'NotAllowedError') {
-        playing = false
-        handlers.error?.('NotAllowedError')
+    audio.play()?.catch(e => handlePlayError(e, { clearStuck: true }))
+
+    // Stuck detection: if onplay never fires within timeout, retry once then error
+    stuckTimer = setTimeout(() => {
+      if (!playing && active && !errored) {
+        if (!stuckRetried) {
```
nwingt: can return early?

composables/use-web-audio-player.ts
```diff
@@ -154,14 +182,26 @@ export function useWebAudioPlayer(): TTSAudioPlayer {
...
+          console.warn(`Audio stuck — retrying playback`)
+          stuckRetried = true
+          audio.load()
+          audio.play()?.catch(e => handlePlayError(e, { clearStuck: true }))
+          stuckTimer = setTimeout(() => {
+            if (!playing && active && !errored) {
```
nwingt: same here

nwingt (approved):

PR #762, opened 2026-03-02
🐛 Fix iOS Safari audio recording by using Minimax-compatible MIME type

williamchong:
iOS Safari's MediaRecorder produces audio/mp4 which decodeAudioData() cannot reliably decode. Now request mp4 format from MediaRecorder when available and pass it through directly to Minimax (which accepts m4a). Only convert to MP3 for unsupported formats like webm/ogg.

### DIFF
--- a/composables/use-audio-recorder.ts
+++ b/composables/use-audio-recorder.ts
@@ -1,3 +1,26 @@
+const PREFERRED_MIME_TYPES = [
+  'audio/mp4',
+  'audio/mpeg',
+]
+
+const PASSTHROUGH_TYPES = new Set([
+  'audio/mpeg',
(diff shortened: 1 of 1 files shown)


composables/use-audio-recorder.ts
```diff
@@ -13,7 +36,7 @@ export function useAudioRecorder(options: {
 
   const isRecording = ref(false)
   const mediaRecorder = ref<MediaRecorder | null>(null)
-  const recordingChunks = ref<Blob[]>([])
+  let recordingChunks: Blob[] = []
```
nwingt, line 39: should put this in `startRecording()`?

nwingt (approved):

PR #778, opened 2026-03-04
👔 Update UI for App Store submit

nwingt:

### DIFF
--- a/pages/account/index.vue
+++ b/pages/account/index.vue
@@ -351,6 +351,7 @@
         </AccountSettingsItem>
 
         <AccountSettingsItem
+          v-if="!isApp"
           icon="i-material-symbols-18-up-rating-outline-rounded"
           :label="$t('account_page_adult_content')"
         >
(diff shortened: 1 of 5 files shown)


pages/account/index.vue
```diff
@@ -540,6 +548,9 @@ const { isApp } = useAppDetection()
 const isAdultContentEnabled = useAdultContentSetting()
 const isAdultContentConfirmOpen = ref(false)
 
+const isPlusFeatureVisible = computed(() => isApp.value || !!user.value?.isLikerPlus)
```
williamchong, line 551: Please confirm if it is `!isApp.value` or `isApp.value`

williamchong (changes requested):

pages/account/index.vue
nwingt: visible in app but show locked icon and disable

williamchong: image

say for this line now for non app non plus user they dont see the whole block?

nwingt: fall to else case, they will see the upgrade to plus button
Screenshot 2026-03-04 at 23 52 42

williamchong (approved):

PR #779, opened 2026-03-04
✨ Add Hong Kong local histories map page at /local-histories/hongkong

williamchong:
image


pages/local-histories/hongkong.vue
nwingt, line 1: if this page intended not using dark mode compatible colors, i suggest to force this page to light mode

nwingt (changes requested):

nwingt:
I think we dont need 4:5 aspect ratio for HK map

nwingt:
The contrast of the text below the title is too low

AuroraHuang22 (approved):

nwingt (approved):

PR #789, opened 2026-03-06
🐛 Fix TTS playback failure on iOS Chrome

williamchong:
Remove idle audio element priming that consumed the user gesture before the real play() call. Short-circuit NotAllowedError to stop immediately instead of retrying without a gesture. Pre-fetch custom voice data in reader pages so the TTS modal can auto-play without an async gap.


nwingt:
seems Copilot's comment is valid?

nwingt (approved):

PR #791, opened 2026-03-09
🐛 Skip yearly upsell for Plus monthly subscribers on free books

williamchong:


nwingt (approved):
is it possible to share logic with another composable to make life easier?

williamchong:
> is it possible to share logic with another composable to make life easier?

another PR maybe, now subscription/upsell is kind of messy

PR #805, opened 2026-03-12
🚸 Adjust zoom functionality for PDF rendering

AuroraHuang22:


components/PDFReader.vue
```diff
@@ -422,13 +423,61 @@ async function loadPDF() {
...
+function clampScale(value: number) {
+  return Math.min(scaleMax, Math.max(scaleMin, value))
+}
+
+async function maybeAutoFitScale() {
+  if (!pdfDocument.value || hasAutoFittedScale.value) return
+  if (scale.value !== 1.0) return
+
+  await nextTick()
+  await new Promise(resolve => requestAnimationFrame(() => resolve(true)))
```
williamchong, line 440: What are we waiting for with `requestAnimationFrame`?

components/PDFReader.vue
```diff
@@ -422,13 +423,61 @@ async function loadPDF() {
...
+    const rightPageNum = dualRightPage.value
+    const pagePromises = [pdfDocument.value.getPage(currentPage.value)]
+    if (rightPageNum) pagePromises.push(pdfDocument.value.getPage(rightPageNum))
+    const pages = await Promise.all(pagePromises)
+
+    const viewports = pages.map(page => page.getViewport({ scale: 1 }))
+    const totalWidth = viewports.reduce((sum, viewport) => sum + viewport.width, 0)
+    const maxHeight = Math.max(...viewports.map(viewport => viewport.height))
+    const effectiveScale = Math.min(containerWidth / totalWidth, containerHeight / maxHeight)
+    const fitScale = clampScale(Math.sqrt(effectiveScale))
```
williamchong: Why `Math.sqrt`?

components/PDFReader.vue
```diff
@@ -422,13 +423,61 @@ async function loadPDF() {
...
   catch (error) {
     emit('error', error as Error)
   }
 }
 
+function clampScale(value: number) {
+  return Math.min(scaleMax, Math.max(scaleMin, value))
+}
+
+async function maybeAutoFitScale() {
```
nwingt:
```suggestion
async function autoZoomToPageIfNeeded() {
```

components/PDFReader.vue
```diff
@@ -258,6 +258,7 @@ const scaleMenuItems = computed<DropdownMenuItem[]>(() => {
 const pdfDocument = shallowRef<PDFDocumentProxy>()
 const renderQueue = ref<(() => Promise<void>)[]>([])
 const isRendering = ref(false)
+const hasAutoFittedScale = ref(false)
```
nwingt:
```suggestion
const hasAutoZoomedToPage = ref(false)
```

components/PDFReader.vue
AuroraHuang22: Waiting for the next paint so layout/size is settled before measuring the container for auto-fit scale.

components/PDFReader.vue
AuroraHuang22: Since both the PDF rendering and the CSS transform use the scale value, we apply Math.sqrt to prevent the double-scaling effect (scale^2).

But we can simplify scaling logic:
[🎨 Use raw scale instead of Math.sqrt](https://github.com/likecoin/3ook-com/pull/805/commits/030041b64f8257215a6c6b9efdbb34c6211a46cf)

nwingt (approved):

williamchong (approved):

PR #808, opened 2026-03-12
🚸 Improve About page value realization for readers 

williamchong:


nwingt (approved):

components/PricingPageContent.vue
```diff
@@ -220,6 +220,14 @@ const selectedPlan = useVModel(props, 'modelValue', emit, {
 
 const isPaidTrial = computed(() => props.trialPeriodDays && props.trialPeriodDays >= PAID_TRIAL_PERIOD_DAYS_THRESHOLD)
 
+const route = useRoute()
+const learnMoreRoute = computed(() => {
+  if (route.name?.toString().startsWith('about')) {
```
nwingt: should use getRouteBaseName for non localized route name?

williamchong: @nwingt the subscription composable overlay has major issues, let me merge first 2 commit and open a new PR

williamchong: https://github.com/likecoin/3ook-com/pull/812

PR #812, opened 2026-03-12
♻️ Split useSubscription into data and modal composables

williamchong:


pages/member.vue
```diff
@@ -36,7 +36,7 @@ definePageMeta({ layout: false })
 
 const localeRoute = useLocaleRoute()
 const getRouteQuery = useRouteQuery()
-const subscription = useSubscription()
+const subscription = useSubscriptionModal()
```
nwingt: this can keep using `useSubscription()` if not using modal?

composables/use-subscription-modal.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  return {
+    ...subscriptionData,
+    isProcessingSubscription,
+
+    openPaywallModal,
+    closePaywallModal,
+    openUpsellPlusModalIfEligible,
+    redirectIfSubscribed,
+    startSubscription,
```
nwingt: should we not expose `redirectIfSubscribed`, `startSubscription` if not used?

nwingt (approved):

PR #825, opened 2026-03-13
[TAS-5425] 🐛 Fix CMS tag items not sorted by staking amount

williamchong:
Staking fetch errors were silently swallowed for CMS tags, causing all items to get 0 totalStaked and preserving Airtable view order. Also fix staking pagination off-by-one that could miss the last page.

### DIFF
--- a/pages/store/index.vue
+++ b/pages/store/index.vue
@@ -901,18 +901,17 @@ async function fetchTags() {
 
 async function fetchTagItems({ isRefresh = false } = {}) {
   const apiSortValue = mapTagIdToAPIStakingSortValue(isStakingTagId.value ? tagId.value : STAKING_TAG_DEFAULT)
-  const fetchPromises = [
-    // NOTE: Fetch staking books for sorting CMS tag items by staking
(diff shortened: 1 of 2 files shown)


pages/store/index.vue
```diff
@@ -901,18 +901,17 @@ async function fetchTags() {
...
+    await bookstoreStore.fetchStakingBooks(apiSortValue, { isRefresh, limit: 100 })
+    return
   }
-  await Promise.all(fetchPromises)
+
+  // Fetch staking books first so CMS tag items can be sorted by staking
+  await bookstoreStore.fetchStakingBooks(apiSortValue, { isRefresh, limit: 100 }).catch((error) => {
+    console.warn('[store] Failed to fetch staking data for CMS tag sorting:', error)
+  })
+  await bookstoreStore.fetchCMSProductsByTagId(tagId.value, { isRefresh })
```
nwingt, line 914: always fetch CMS products even under staking tags?

williamchong: staking tag returned at line 907

PR #826, opened 2026-03-13
[TAS-5409] 🚀 Add dual-TTL server cache for Airtable CMS store APIs

williamchong:
Cache first-page records (1 day) and offset cursors (2 min) separately to avoid Airtable rate limits while handling cursor expiry gracefully. Paginated requests always bypass cache. Client detects missing offset on full pages and re-fetches page 1 for a fresh cursor without flash.

### DIFF
--- a/stores/bookstore.ts
+++ b/stores/bookstore.ts
@@ -70,14 +70,14 @@ export const useBookstoreStore = defineStore('bookstore', () => {
 
     // If offset expired but more pages likely exist, re-fetch page 1 to get fresh offset
     const state = bookstoreCMSProductsByTagIdMap.value[tagId]
-    const needsOffsetRefresh = !isRefresh && state?.hasFetched && !state?.offset && state?.mayHaveMore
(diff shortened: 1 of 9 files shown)


stores/bookstore.ts
```diff
@@ -62,7 +67,17 @@ export const useBookstoreStore = defineStore('bookstore', () => {
     if (bookstoreCMSProductsByTagIdMap.value[tagId]?.isFetching) {
       return
     }
-    if (!bookstoreCMSProductsByTagIdMap.value[tagId] || isRefresh) {
+
+    // If offset expired but more pages likely exist, re-fetch page 1 to get fresh offset
+    const state = bookstoreCMSProductsByTagIdMap.value[tagId]
+    const needsOffsetRefresh = !isRefresh && state?.hasFetched && !state?.offset && state?.mayHaveMore
```
nwingt:
```suggestion
    const shouldRefreshOffset = !isRefresh && state?.hasFetched && !state?.offset && state?.mayHaveMore
```

nwingt (approved):

PR #831, opened 2026-03-16
✨ Add Cantonese TTS model selection (口語/書面語)

williamchong:
Allow users to switch between Minimax speech-2.6 (colloquial) and speech-2.8 (written) for zh-HK voices. Stored as a synced user setting, passed as minimax_model query param to the TTS endpoint, and guarded server-side in getMinimaxModel to only apply for zh-HK.


pages/account/index.vue
```diff
@@ -566,6 +586,11 @@ function confirmAdultContent() {
 }
 
 const { customVoice, hasCustomVoice, fetchCustomVoice } = useCustomVoice()
+const ttsCantoneseModel = useTTSCantoneseModel()
+const cantoneseModelOptions = [
+  { label: $t('tts_cantonese_model_option_colloquial'), value: '2.6' },
+  { label: $t('tts_cantonese_model_option_written'), value: '2.8' },
+]
```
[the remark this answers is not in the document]
nwingt: @williamchong Please use computed

pages/account/index.vue
```diff
@@ -310,6 +310,26 @@
...
+          <template #right>
+            <UButtonGroup>
+              <UButton
+                v-for="option in cantoneseModelOptions"
+                :key="option.value"
+                :label="option.label"
+                :variant="ttsCantoneseModel === option.value ? 'solid' : 'outline'"
+                color="neutral"
+                size="xs"
+                @click="ttsCantoneseModel = option.value"
```
nwingt: @williamchong Please fix the type
btw should we add event too?

server/api/reader/tts.get.ts
```diff
@@ -114,7 +114,7 @@ export default defineEventHandler(async (event) => {
     })
   }
   const config = useRuntimeConfig()
-  const { text, language, voice_id: voiceId, blocking, nft_class_id: nftClassId } = await getValidatedQuery(event, createValidator(TTSQuerySchema))
+  const { text, language, voice_id: voiceId, blocking, nft_class_id: nftClassId, minimax_model: minimaxModel } = await getValidatedQuery(event, createValidator(TTSQuerySchema))
```
nwingt: spread with multiline?

server/utils/tts-minimax.ts
```diff
@@ -40,7 +40,10 @@ export function getTTSPronunciationDictionary(language: string) {
   }
 }
 
-export function getMinimaxModel(customMiniMaxVoiceId?: string, language?: string): string {
+export function getMinimaxModel(customMiniMaxVoiceId?: string, language?: string, preferredModel?: string): string {
```
nwingt: consider using object params

nwingt (changes requested):

nwingt (approved):

PR #834, opened 2026-03-17
💬 Rename 粵語語音模型 to 粵語風格 and default to 書面語

williamchong:


server/utils/tts-minimax.ts
```diff
@@ -47,7 +47,7 @@ export function getMinimaxModel(options: {
 } = {}): string {
   const { customVoiceId, language, preferredModel } = options
   if (language === 'zh-HK') {
-    return preferredModel === '2.8' ? 'speech-2.8-hd' : 'speech-2.6-hd'
+    return preferredModel === '2.6' ? 'speech-2.6-hd' : 'speech-2.8-hd'
```
nwingt, line 50: what is the difference?

nwingt (approved):

server/utils/tts-minimax.ts
williamchong: syntactically we prefer 2.8 if nothing is set, but practically same

PR #839, opened 2026-03-17
[TAS-5413] ✨ Add sponsored transactions for Magic users via EIP-7702

williamchong:
Introduce gasless contract writes for Magic Link users using Alchemy's Account Kit with EIP-7702 delegation and paymaster sponsorship.

- Add useContractWrite() wrapper that routes Magic users through sponsored path with graceful fallback to direct transactions
- Add useSponsoredTransaction() with lazy-loaded Alchemy SDK, cached smart wallet client, and dedup for concurrent calls
- Replace useWriteContract() with useContractWrite() in all 5 contract composables (likecoin, veLike, veLikeReward, collective, nft)
- Add ALCHEMY_GAS_POLICY_ID env var and sort env files alphabetically


composables/use-sponsored-transaction.ts
```diff
@@ -0,0 +1,176 @@
...
+  cachedMagicInstance = null
+  pendingClientCreation = null
+}
+/* eslint-enable @typescript-eslint/no-explicit-any */
+
+export function useSponsoredTransaction() {
+  const accountStore = useAccountStore()
+  const config = useRuntimeConfig()
+  const { $wagmiConfig } = useNuxtApp()
+  const alchemyApiKey = computed(() => (config.public.customRpcUrl as string)?.split('/v2/')[1])
```
nwingt:
```suggestion
  const alchemyAPIKey = computed(() => (config.public.customRpcUrl as string)?.split('/v2/')[1])
```

nuxt.config.ts
```diff
@@ -76,6 +76,7 @@ export default defineNuxtConfig({
       airtableCMSPublicationsTableId: process.env.AIRTABLE_CMS_PUBLICATIONS_TABLE_ID,
       airtableCMSTagsTableId: process.env.AIRTABLE_CMS_TAGS_TABLE_ID,
       cacheKeyPrefix: '3ook',
+      alchemyGasPolicyId: process.env.ALCHEMY_GAS_POLICY_ID,
```
nwingt:
```suggestion
      alchemyGasPolicyId: process.env.ALCHEMY_GAS_POLICY_ID,
      cacheKeyPrefix: '3ook',
```

nwingt (approved):

PR #849, opened 2026-03-19
[TAS-5471] 🚀 Merge short TTS segments for better buffering and streaming

williamchong:
Add MIN_SEGMENT_LENGTH threshold (15 chars) to avoid producing very short segments that each trigger a separate TTS API call. Short fragments are now merged with adjacent segments at two levels:
- mergeParts: absorbs short trailing text and sub-threshold parts
- mergeShortTTSSegments: cross-element merge within the same section

### DIFF
--- a/pages/reader/epub.vue
+++ b/pages/reader/epub.vue
@@ -1000,7 +1000,7 @@ async function extractTTSSegments(book: Book) {
     }
   }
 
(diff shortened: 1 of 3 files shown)


nwingt (approved):
Trust William

nwingt (approved):
Trust Copilot & William

PR #856, opened 2026-03-20
🚸 Sort CMS tag items per fetch batch to preserve scroll position

williamchong:
Global staking sort caused newly fetched items to be interleaved into already-scrolled positions. Sort each batch independently so new items always appear below the current scroll position.

### DIFF
--- a/pages/store/index.vue
+++ b/pages/store/index.vue
@@ -672,14 +672,6 @@ const cmsProducts = computed<BookstoreItemList>(() => {
     }
   })
 
-  if (tagId.value !== 'latest') {
-    items.sort((a, b) => {
-      const aTotalStaked = a.totalStaked ?? 0n
-      const bTotalStaked = b.totalStaked ?? 0n
-      return Number(bTotalStaked - aTotalStaked)
-    })
-  }
-
   return {
     ...listingProducts,
     items,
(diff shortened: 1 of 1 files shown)


pages/store/index.vue
```diff
@@ -911,7 +903,30 @@ async function fetchTagItems({ isRefresh = false } = {}) {
...
+  // if the store replaced it (e.g. on refresh or expired-offset retry)
+  const itemsBefore = bookstoreStore.bookstoreCMSProductsByTagIdMap[tagId.value]?.items
+  const countBefore = isRefresh ? 0 : (itemsBefore?.length ?? 0)
   await bookstoreStore.fetchCMSProductsByTagId(tagId.value, { isRefresh })
+
+  // Sort only the new batch by staking amount (skip 'latest' which preserves Airtable order)
+  if (tagId.value !== 'latest') {
+    const items = bookstoreStore.bookstoreCMSProductsByTagIdMap[tagId.value]?.items
+    // If the array was replaced (refresh or offset-refresh), sort from 0
+    const fromIndex = (items === itemsBefore) ? countBefore : 0
```
nwingt:
```suggestion
    const sortingFromIndex = (items === itemsBefore) ? countBefore : 0
```

nwingt (approved):

PR #858, opened 2026-03-20
[TAS-5474] 🐛 Normalize fullwidth ASCII characters in TTS text sanitization

williamchong:


nwingt (approved):

utils/tts.ts
```diff
@@ -1,11 +1,13 @@
 export function sanitizeTTSText(text: string): string {
   if (!text) return ''
   return text
+    .replace(/[\uFF01-\uFF5E]/g, c =>
+      String.fromCharCode(c.charCodeAt(0) - 0xFEE0))
```
nwingt, line 6: Maybe add a comment for what its doing

nwingt (approved):

PR #864, opened 2026-03-23
[TAS-5551] 🔥 Remove Cantonese TTS model toggle and always use speech-2.8-hd

williamchong:
Remove the 口語/書面語 toggle from account settings and simplify the TTS pipeline to always use speech-2.8-hd for Cantonese (zh-HK), eliminating the minimax_model query parameter and preferredModel option throughout the stack.


nwingt (approved):
Should we just hide the entry?

williamchong:
> Should we just hide the entry?

Dont want existing config to mess up

PR #887, opened 2026-03-27
♻️ Refactor and add back buttons and map previews to local histories pages

williamchong:
image


nwingt (changes requested):
Please fix the hover buttons style

PR #906, opened 2026-04-01
[TAS-5503] 👔 Hide Intercom UI for app

nwingt:


williamchong (commented):
Do we want to just disable intercom completely if in app? Or just this icon and some intercom function would be still working?

nwingt:
@williamchong I think UI wise is enough, whats your thought?

ckxpress:
we might come back to enable it later. it's low priority and may not be too near future, so i simply requested to remove it.

for now, most important is, i think, don't give it a chance to show accidentally and ruin our review, etc

plugins/intercom.client.ts
```diff
@@ -0,0 +1,16 @@
...
+  }
+
+  // Hide Intercom UI elements via CSS in case it loads later
+  const styleId = 'intercom-hide-style'
+  if (!document.getElementById(styleId)) {
+    const style = document.createElement('style')
+    style.id = styleId
+    style.textContent = '#intercom-container, .intercom-lightweight-app, .intercom-namespace { display: none !important; }'
+    document.head.appendChild(style)
+  }
```
williamchong, line 15: Claude code say we can hook into intercom script, please try
```suggestion
  const intercom = useScriptIntercom()
  intercom.onLoaded(({ Intercom }) => {
    Intercom('update', { hide_default_launcher: true })
  })
```

williamchong (changes requested):

plugins/intercom.client.ts
nwingt: tried. it doesnt work

williamchong (approved):

PR #907, opened 2026-04-01
[TAS-5596] ✨ Add writing mode options for EPUB reader and update settings

AuroraHuang22:
截圖 2026-04-01 21 41 50


pages/reader/epub.vue
```diff
@@ -204,6 +204,32 @@
...
                   </div>
+                  <div class="flex-1">
+                    <p
+                      class="block text-sm font-medium mb-2"
+                      v-text="$t('reader_display_orientation_label')"
+                    />
+                    <div class="grid grid-cols-2 gap-2">
+                      <UButton
+                        :variant="writingMode === EPUB_WRITING_MODES.horizontal ? 'solid' : 'outline'"
+                        :label="$t('reader_display_orientation_horizontal')"
```
nwingt: text should be align to center
```suggestion
                        :label="$t('reader_display_orientation_horizontal')"
                        block
```

pages/reader/epub.vue
```diff
@@ -204,6 +204,32 @@
...
+                        @click="setWritingMode(EPUB_WRITING_MODES.horizontal)"
+                      />
+                      <UButton
+                        :variant="writingMode === EPUB_WRITING_MODES.vertical ? 'solid' : 'outline'"
+                        :label="$t('reader_display_orientation_vertical')"
```
nwingt: text should be align to center
```suggestion
                        :label="$t('reader_display_orientation_vertical')"
                        block
```

but I think use icon is better

`i-material-symbols-text-rotate-vertical-rounded`

Image

pages/reader/epub.vue
nwingt: but i think use icon is better

`i-material-symbols-text-rotation-none-rounded`

Image

pages/reader/epub.vue
```diff
@@ -1105,6 +1200,57 @@ function decreaseLineHeight() {
...
+  useLogEvent('reader_setting_changed', { nft_class_id: nftClassId.value, setting: 'writing_mode', value: mode })
+}
+
+async function restoreDefaultDisplayOptions() {
+  const previousFontSize = fontSize.value
+  const previousLineHeight = lineHeight.value
+  const previousWritingMode = writingMode.value
+
+  fontSize.value = FONT_SIZE_OPTIONS[DEFAULT_FONT_SIZE_INDEX]
+  lineHeight.value = DEFAULT_LINE_HEIGHT
```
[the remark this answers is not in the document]
nwingt: @AuroraHuang22 seems true

pages/reader/epub.vue
```diff
@@ -266,7 +292,7 @@
                 'flex-1',
                 'h-full',
                 'cursor-pointer',
-                { 'opacity-0 pointer-events-none': isRightToLeft ? isAtFirstPage : isAtLastPage },
+                { 'opacity-0 pointer-events-none': isDefaultRightToLeft ? isAtFirstPage : isAtLastPage },
```
nwingt: i think we should keep isRightToLeft for reversing page navigation according to writing mode

nwingt (changes requested):

nwingt (approved):

pages/reader/epub.vue
```diff
@@ -771,7 +868,6 @@ async function loadEPub() {
 
   rendition.value.on('rendered', (section: Section, view: EpubView) => {
     currentSectionIndex.value = section.index ?? 0
-    isRightToLeft.value = view.settings.direction === 'rtl'
```
williamchong, line 774: Instead of setting `const DEFAULT_WRITING_MODE = EPUB_WRITING_MODES.horizontal`, can we read the existing direction/writing mode from epub?

williamchong (changes requested):

PR #909, opened 2026-04-01
[TAS-5579] 🚸 Make register CTA mobile-only and always show TTS samples on mobile

williamchong:
image

Use CSS responsive classes instead of JS media queries to avoid SSR hydration mismatch. Add pricing-mobile slot for mobile-specific content override with display:contents fallback on desktop.

### DIFF
--- a/components/PricingPageContent.vue
+++ b/components/PricingPageContent.vue
@@ -119,37 +119,47 @@
           :description="campaignContent?.description"
           :is-compact="isShowTTSSamples"
         />
-        <TTSSamplesSection v-if="isShowTTSSamples" />
+        <TTSSamplesSection :class="{ 'laptop:hidden': !isShowTTSSamples }" />
 
         <div class="flex flex-col w-full mt-6 laptop:mt-8">
-          <PricingLimitedOfferAlert
-            v-if="!isApp"
-            :is-hidden="!isPaidTrial"
+          <div
+            v-if="$slots['pricing-mobile']"
+            class="laptop:hidden"
           >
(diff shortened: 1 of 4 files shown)


pages/member.vue
```diff
@@ -22,6 +22,26 @@
...
     </template>
+
+    <template
+      v-if="!hasLoggedIn"
+      #pricing-mobile
+    >
+      <div class="flex flex-col items-center gap-3">
+        <span class="text-sm text-theme-black/60 dark:text-theme-white/60">
+          {{ $t('pricing_page_login_cta_description') }}
+        </span>
```
nwingt:
```suggestion
        <span
          class="text-sm text-muted"
          v-text="$t('pricing_page_login_cta_description')"
        />
```

components/PricingPageContent.vue
```diff
@@ -119,37 +119,47 @@
...
-        <TTSSamplesSection v-if="isShowTTSSamples" />
+        <TTSSamplesSection :class="{ 'laptop:hidden': !isShowTTSSamples }" />
 
         <div class="flex flex-col w-full mt-6 laptop:mt-8">
-          <PricingLimitedOfferAlert
-            v-if="!isApp"
-            :is-hidden="!isPaidTrial"
+          <div
+            v-if="$slots['pricing-mobile']"
+            class="laptop:hidden"
```
nwingt:
```suggestion
            class="contents laptop:hidden"
```

nwingt (changes requested):

nwingt (approved):

PR #910, opened 2026-04-01
[TAS-5601] 🚸 Add plus to TTS tag on product page

williamchong:
image


nwingt:
too dimmed in dark mode

williamchong:
> too dimmed in dark mode 

Updated to just use primary for dark mode

nwingt:
@williamchong I have made some UI updates https://github.com/williamchong/liker-land-v3/pull/17

nwingt (approved):

PR #917, opened 2026-04-02
⚗️ Gate mobile register CTA and TTS samples behind A/B test

williamchong:
Wrap the mobile-only register CTA and force-show TTS samples behavior behind a PostHog experiment (pricing-page-mobile-register-cta) instead of deploying directly to all users.

### DIFF
--- a/components/PricingPageContent.vue
+++ b/components/PricingPageContent.vue
@@ -120,7 +120,7 @@
           :is-compact="isShowTTSSamples"
         />
         <TTSSamplesSection
-          v-if="isShowTTSSamples || props.forceShowTTSOnMobile"
+          v-if="isShowTTSSamples || props.mustShowTtsOnMobile"
           :class="{ 'laptop:hidden': !isShowTTSSamples }"
         />
 
--- a/components/PricingPageContent.vue
+++ b/components/PricingPageContent.vue
@@ -119,10 +119,7 @@
           :description="campaignContent?.description"
           :is-compact="isShowTTSSamples"
         />
-        <TTSSamplesSection
-          v-if="isShowTTSSamples || props.mustShowTtsOnMobile"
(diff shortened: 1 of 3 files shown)


components/PricingPageContent.vue
```diff
@@ -119,7 +119,10 @@
           :description="campaignContent?.description"
           :is-compact="isShowTTSSamples"
         />
-        <TTSSamplesSection :class="{ 'laptop:hidden': !isShowTTSSamples }" />
+        <TTSSamplesSection
+          v-if="isShowTTSSamples || props.forceShowTTSOnMobile"
```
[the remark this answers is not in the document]
nwingt: should add `props.forceShowTTSOnMobile` to the computed `isShowTTSSamples` instead of here?

pages/member.vue
```diff
@@ -9,6 +9,7 @@
     utm-source="website"
     utm-medium="web"
     :coupon="coupon"
+    :force-show-t-t-s-on-mobile="mobileRegisterCTATest.isVariant('test')"
```
nwingt: this really hurt my eyes 😵

pages/member.vue
```diff
@@ -24,7 +25,7 @@
     </template>
 
     <template
-      v-if="!hasLoggedIn"
+      v-if="!hasLoggedIn && mobileRegisterCTATest.isVariant('test')"
```
nwingt: should create a computed for `mobileRegisterCTATest.isVariant('test')`?

pages/member.vue
williamchong: either `force-show-t-t-s-on-mobile` or `forceShowTtsOnMobile` (not TTS), see which one you hate more

pages/member.vue
williamchong: Updated

nwingt (approved):

PR #924, opened 2026-04-07
✨ Persist TTS playback state in URL query param

williamchong:
Sync ?tts=1 with player open/close so TTS auto-resumes on page load in both EPUB and PDF readers.

### DIFF
--- a/composables/use-tts-query-param.ts
+++ b/composables/use-tts-query-param.ts
@@ -1,0 +1,19 @@
+export function useTTSQueryParam() {
+  const route = useRoute()
+  const router = useRouter()
+
+  function setTTSQueryParam(value: boolean) {
+    const alreadySet = !!route.query.tts
+    if (value === alreadySet) return
+    const query = { ...route.query }
+    if (value) {
+      query.tts = '1'
+    }
+    else {
+      delete query.tts
+    }
+    router.replace({ query })
+  }
+
+  return { setTTSQueryParam }
+}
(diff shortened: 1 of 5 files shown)


composables/use-tts-query-param.ts
```diff
@@ -0,0 +1,19 @@
+export function useTTSQueryParam() {
+  const route = useRoute()
+  const router = useRouter()
+
+  function setTTSQueryParam(value: boolean) {
+    const alreadySet = !!route.query.tts
+    if (value === alreadySet) return
```
nwingt:
```suggestion
    const hasAlreadySet = !!route.query.tts
    if (value === hasAlreadySet) return
```

pages/reader/epub.vue
```diff
@@ -917,15 +920,35 @@ async function loadEPub() {
...
   })
 
-  if (shouldShowTTSTryModal.value && !bookInfo.isAudioHidden.value) {
-    showTTSTryModal({
-      nftClassId: nftClassId.value,
-      onVoiceSelected: () => {
-        onClickTTSPlay()
-      },
-    })
+  if (route.query.tts) {
```
nwingt: should we use a computed that expose this from `useTTSQueryParam()`?

pages/reader/pdf.vue
```diff
@@ -146,6 +149,15 @@ function handlePDFLoaded(pdfDocument: PDFDocumentProxy) {
   updatePDFProgress(currentPageIndex.value)
   // Release the ArrayBuffer — PDF.js has its own internal copy
   fileBuffer.value = null
+
+  if (route.query.tts) {
```
nwingt: same here

nwingt: same here

nwingt (approved):

PR #939, opened 2026-04-10
🎨 Clean up pricing components

nwingt:
Separate limited offer UI from PricingPlanSelect, removed `Limited Offer` badge as it has already appeared in header


components/PricingPageContent.vue
```diff
@@ -131,14 +131,41 @@
 
           <div :class="$slots['pricing-mobile'] ? 'hidden laptop:contents' : undefined">
             <slot name="pricing">
-              <PricingLimitedOfferAlert
-                v-if="!isApp"
-                :is-hidden="!isPaidTrial"
+              <div
+                v-if="!isApp || isPaidTrial"
```
williamchong: Should this be &&?

williamchong (changes requested):

components/PricingPageContent.vue
nwingt: outdated

williamchong (approved):

PR #944, opened 2026-04-11
⚡️ Defer Meta Pixel composable lookup to mapped events

williamchong:
useScriptMetaPixel() was called on every useLogEvent invocation, including hot-path reader events (arrow-key, swipe). Only 7 events map to fbq, so hoist the lookup inside the mapping-table branch.


nwingt (approved):
Should use ⚡️ instead of 🚀?

PR #959, opened 2026-04-15
🚸 Use danger color for delete uploaded book

nwingt:


williamchong (approved):
Maybe add a confirm modal

PR #960, opened 2026-04-15
🐛 Fix upload book item menu title in mobile

nwingt:


williamchong (approved):

ckxpress:
actually, let's make upload book a computer only feature. avoid app/mobile to access.

PR #966, opened 2026-04-15
📈 Instrument TTS trial usage for conversion analysis

williamchong:
Adds an internal usage endpoint plus richer event properties (is_liker_plus_at_event_time, cumulative_chars_used, chars_remaining, book_language, voice_id) so the TTS free→Plus funnel can be measured before any product change lands. Fires tts_trial_exhausted on the paywall path and ships an optimistic client-side char decrement so a future trial chip has accurate state without extra round-trips.

### DIFF
--- a/composables/use-text-to-speech.ts
+++ b/composables/use-text-to-speech.ts
@@ -46,14 +46,26 @@ export function useTextToSpeech(options: TTSOptions) {
     setTTSLanguageVoice,
   } = useTTSVoice({ bookLanguage, customVoice: options.customVoice })
 
-  function buildTTSEventBase(extras: Record<string, unknown> = {}) {
+  function parseLanguageVoice(languageVoice: string) {
+    const [language = '', ...voiceIdParts] = languageVoice.split('_')
+    return { language, voiceId: voiceIdParts.join('_') }
+  }
+
+  function buildTTSEventPayload(extras: Record<string, unknown> = {}) {
+    const languageVoice = ttsLanguageVoice.value || ''
+    const { voiceId } = parseLanguageVoice(languageVoice)
     return {
       nft_class_id: nftClassId,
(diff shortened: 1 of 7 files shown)


composables/use-text-to-speech.ts
```diff
@@ -43,6 +46,18 @@ export function useTextToSpeech(options: TTSOptions) {
     setTTSLanguageVoice,
   } = useTTSVoice({ bookLanguage, customVoice: options.customVoice })
 
+  function buildTTSEventBase(extras: Record<string, unknown> = {}) {
```
```suggestion
  function buildTTSEventPayload(extras: Record<string, unknown> = {}) {
```

composables/use-tts-trial-usage.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const charactersUsed = computed(() => data.value?.charactersUsed ?? 0)
+  const limit = computed(() => data.value?.limit ?? 0)
+  const charactersRemaining = computed(() => data.value?.charactersRemaining ?? null)
+  const isExhausted = computed(() => data.value?.isExhausted ?? false)
+
+  function fetchUsage(): Promise<void> {
+    if (!hasLoggedIn.value) return Promise.resolve()
+    if (user.value?.isLikerPlus) return Promise.resolve()
+    if (data.value !== null) return Promise.resolve()
```
nwingt:
```suggestion
    if (!hasLoggedIn.value || user.value?.isLikerPlus || data.value !== null) {
      return Promise.resolve()
    }
```

server/api/reader/tts/usage.get.ts
nwingt, line 1: should include trial in the path?

williamchong: Not limited to trial only in the future

nwingt (approved):

PR #967, opened 2026-04-15
✨ Add TTS trial usage chip in reader player

williamchong:
Shows free users their remaining character budget inline with the TTS controls so the wall is no longer invisible until the first failed play. Clicking the chip opens the paywall with utm_medium=tts_chip so the lift over the 12.45% baseline can be attributed cleanly.
image

image


composables/use-text-to-speech.ts
```diff
@@ -502,20 +508,12 @@ export function useTextToSpeech(options: TTSOptions) {
     }
   }
 
-  // Estimate only: the server charges on cache miss, so we dedupe by the
-  // same shape (language+voice+text) that determines its cache key.
-  const optimisticallyCountedRequests = new Set<string>()
-
+  // Custom voice is gated to Plus; the server rejects non-Plus requests
+  // before charging, so decrementing here would drift the local counter.
```
nwingt: is this comment referring the if below?

nwingt (commented):
Please review too
https://github.com/williamchong/liker-land-v3/pull/20

composables/use-text-to-speech.ts
williamchong: Yes it is explaining `if (ttsLanguageVoice.value === 'custom') return`

nwingt (approved):

PR #971, opened 2026-04-16
💄 Use plus icon for add-to-bookshelf button

williamchong:
Swap the hollow heart for a plus sign when a book isn't yet in the reader's bookshelf, so the action reads as "add" rather than "favorite".
image


nwingt (commented):
please use i-material-symbols-add-2-rounded

nwingt (approved):

PR #979, opened 2026-04-17
[TAS-5535] ✨ Add DNF book state

nwingt:
image

image

image

image

image


stores/bookshelf.ts
```diff
@@ -212,6 +216,24 @@ export const useBookshelfStore = defineStore('bookshelf', () => {
...
+    bookSettingsStore.queueUpdate(normalizedNFTClassId, 'didNotFinishAt', Date.now())
+    bookSettingsStore.queueUpdate(normalizedNFTClassId, 'completedAt', null)
+  }
+
+  function markBookAsReading(nftClassId: string) {
+    const normalizedNFTClassId = nftClassId.toLowerCase()
+    bookSettingsStore.queueUpdate(normalizedNFTClassId, 'completedAt', null)
+    bookSettingsStore.queueUpdate(normalizedNFTClassId, 'didNotFinishAt', null)
+  }
+
```
williamchong, line 236: Seems more robust to set time at backend using server time, instead of passing now timestamp from frontend

williamchong (approved):

PR #984, opened 2026-04-19
📈 Add TTS player timing events

williamchong:


composables/use-text-to-speech.ts
```diff
@@ -56,6 +56,25 @@ export function useTextToSpeech(options: TTSOptions) {
...
+  const { isApp, isIOS, isAndroid } = useAppDetection()
+  const isNativeBridge = computed(
+    () => isApp.value && isNativeWebView(),
+  )
+  const appPlatform = computed<'ios' | 'android' | 'web'>(() => {
+    if (!isNativeBridge.value) return 'web'
+    if (isIOS.value) return 'ios'
+    if (isAndroid.value) return 'android'
+    return 'web'
+  })
```
nwingt: should we put these into `useAppDetection()`?

composables/use-text-to-speech.ts
```diff
@@ -132,6 +147,50 @@ export function useTextToSpeech(options: TTSOptions) {
...
+  function clearSegmentLoadTimer() {
+    segmentLoadStart = null
+  }
+
+  const MEDIA_ERROR_NAMES: Record<number, string> = {
+    1: 'MEDIA_ERR_ABORTED',
+    2: 'MEDIA_ERR_NETWORK',
+    3: 'MEDIA_ERR_DECODE',
+    4: 'MEDIA_ERR_SRC_NOT_SUPPORTED',
+  }
```
nwingt: should we put this outside of this function?

composables/use-text-to-speech.ts
```diff
@@ -132,6 +147,50 @@ export function useTextToSpeech(options: TTSOptions) {
...
+
+  function formatTTSError(error: string | Event | MediaError): string {
+    if (typeof error === 'string') return error
+    if (error instanceof MediaError) return MEDIA_ERROR_NAMES[error.code] || `MEDIA_ERR_${error.code}`
+    if (error instanceof Event) return error.type || 'event'
+    return 'unknown'
+  }
```
```suggestion
  }

```

nwingt (approved):

PR #992, opened 2026-04-21
📈 Track TTS browser cache hit/miss per segment

williamchong:
had_buffering conflates a browser <audio> stall (web) with a fresh playTrack call (native), so it can't answer "was this segment cached". Classify each web load via PerformanceResourceTiming.transferSize and emit a cache_status property on tts_segment_loaded / tts_segment_load_failed; native bridge reports 'native' since browser cache doesn't apply.

### DIFF
--- a/composables/use-native-audio-player.ts
+++ b/composables/use-native-audio-player.ts
@@ -108,6 +108,10 @@ export function useNativeAudioPlayer(isActive: Ref<boolean | undefined>): TTSAud
     return false
   }
 
+  function getCurrentURL(): string {
+    return ''
+  }
+
   return {
     load,
     resume,
@@ -118,6 +122,7 @@ export function useNativeAudioPlayer(isActive: Ref<boolean | undefined>): TTSAud
     seek,
     getPosition,
     wasInterruptedByBackground,
+    getCurrentURL,
     on,
   }
 }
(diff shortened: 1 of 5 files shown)


composables/use-native-audio-player.ts
```diff
@@ -108,6 +108,10 @@ export function useNativeAudioPlayer(isActive: Ref<boolean | undefined>): TTSAud
     return false
   }
 
+  function getCurrentURL(): string {
+    return ''
```
nwingt, line 113: maybe add a comment for stub

utils/resource-timing.ts
```diff
@@ -0,0 +1,40 @@
+export type TTSCacheStatus = 'hit' | 'miss' | 'unknown'
+
+const TTS_URL_PATTERN = /\/api\/reader\/tts(?:\?|$)/
+const MAX_TRACKED_URLS = 200
+
+const latestByUrl = new Map<string, PerformanceResourceTiming>()
```
nwingt:
```suggestion
const latestByURL = new Map<string, PerformanceResourceTiming>()
```

utils/resource-timing.ts
```diff
@@ -0,0 +1,40 @@
...
+  try {
+    nextObserver.observe({ type: 'resource', buffered: true })
+    observer = nextObserver
+  }
+  catch {
+    // Environment accepts PerformanceObserver but not this entry type / buffered
+  }
+}
+
+export function classifyTTSCacheStatus(audioUrl: string): TTSCacheStatus {
```
nwingt:
```suggestion
export function classifyTTSCacheStatus(audioURL: string): TTSCacheStatus {
```

PR #997, opened 2026-04-22
🔧 Add robots.txt crawl rules and sitemap reference

williamchong:
Previously a 1-byte placeholder, which implicitly allowed crawlers on every path including /checkout, /account, /shelf, /reader, and claim flows. Add explicit Disallow rules for authenticated/transactional paths and point crawlers at the existing @nuxtjs/sitemap output.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,1 +1,26 @@
+# 3ook.com — robots.txt
+# Decentralized bookstore and AI reading companion
+# https://3ook.com
 
+User-agent: *
+Allow: /
+Disallow: /api/
+Disallow: /checkout
+Disallow: /account
+Disallow: /shelf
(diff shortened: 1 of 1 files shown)


public/robots.txt
```diff
@@ -1 +1,28 @@
...
+# https://3ook.com
 
+User-agent: *
+Allow: /
+Disallow: /api/
+Disallow: /checkout
+Disallow: /account
+Disallow: /shelf
+Disallow: /reader
+Disallow: /local-histories
```
nwingt: wonder why disallow `/local-histories`?

nwingt (approved):

public/robots.txt
williamchong: actually should allow

PR #1001, opened 2026-04-22
💄 Unify account settings rows with current-value + action button

williamchong:
image

Match the 私人聲優 row pattern for 語言, 付款貨幣, and 外觀: show the current choice in the second column and a neutral outline button in the third column that opens a UDropdownMenu. Also rename 外觀顏色 to 外觀 (Color Mode → Appearance).


components/ColorModeSwitcher.vue
```diff
@@ -1,47 +1,30 @@
...
-])
-
-const selectedColorMode = computed(() => preference.value)
-
-function handleColorModeChange(value: string) {
-  preference.value = value as ColorMode
-  useLogEvent('color_mode_switch', { colorMode: value })
-}
+const menuItems = computed<DropdownMenuItem[]>(() =>
+  getColorModeOptions($t).map(option => ({
```
nwingt: should we follow `const { locales, setLocale } = useAutoLocale()` structure?
put `colorModeOptions` in `useColorModeSync`

components/CurrencySwitcher.vue
```diff
@@ -1,32 +1,30 @@
...
-  { label: '🇹🇼 TWD', value: 'twd' },
-  { label: '🇺🇸 USD', value: 'usd' },
-])
-
-function handleCurrencyChange(value: string) {
-  setCurrency(value as PaymentCurrency)
-  useLogEvent('currency_switch', { currency: value })
-}
+const menuItems = computed<DropdownMenuItem[]>(() =>
+  getCurrencyOptions($t).map(option => ({
```
nwingt: same here, should we put `currencyOptions` in `usePaymentCurrency`?

nwingt (approved):

PR #1012, opened 2026-04-23
🐛 Validate persisted TTS index against current page before resuming

williamchong:
The persisted activeTTSElementIndex outlives reader navigations that don't clear it (in-book search, annotation jump), so clicking TTS play after such a jump yanked the reader back to the stale segment's page. Let openPlayer compare the stored segment CFI against the current page range and fall through to the CFI lookup when it no longer lives on the visible page.

### DIFF
--- a/composables/use-tts-player-modal.ts
+++ b/composables/use-tts-player-modal.ts
@@ -59,13 +59,36 @@ export function useTTSPlayerModal(options: TTSPlayerOptions) {
     ttsIndex,
     sectionIndex,
     cfi,
+    pageEndCFI,
     ...props
   }: {
     ttsIndex?: number
     sectionIndex?: number
     cfi?: string
+    pageEndCFI?: string
   } & TTSPlayerModalProps = {}) {
-    if (ttsIndex !== undefined) {
+    const epubCFI = new EpubCFI()
+    // A persisted ttsIndex can outlive a reader navigation that didn't clear
+    // it (search/annotation jump), so only honor it when the page range says
+    // its segment is still on-screen. Otherwise fall through to the cfi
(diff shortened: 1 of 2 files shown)


composables/use-tts-player-modal.ts
```diff
@@ -59,13 +59,29 @@ export function useTTSPlayerModal(options: TTSPlayerOptions) {
...
     cfi?: string
+    pageEndCfi?: string
   } & TTSPlayerModalProps = {}) {
-    if (ttsIndex !== undefined) {
+    const epubCFI = new EpubCFI()
+    // A persisted ttsIndex can outlive a reader navigation that didn't clear
+    // it (search/annotation jump), so only honor it when the page range says
+    // its segment is still on-screen. Otherwise fall through to the cfi
+    // lookup so TTS starts where the user is actually reading.
+    const storedSegmentCfi = ttsIndex !== undefined
```
nwingt:
```suggestion
    const storedSegmentCFI = ttsIndex !== undefined
```

composables/use-tts-player-modal.ts
```diff
@@ -59,13 +59,29 @@ export function useTTSPlayerModal(options: TTSPlayerOptions) {
     ttsIndex,
     sectionIndex,
     cfi,
+    pageEndCfi,
     ...props
   }: {
     ttsIndex?: number
     sectionIndex?: number
     cfi?: string
+    pageEndCfi?: string
```
nwingt:
```suggestion
    pageEndCFI,
...props
}: {
ttsIndex?: number
sectionIndex?: number
cfi?: string
    pageEndCFI?: string
```

PR #1015, opened 2026-04-24
[TAS-5731] 💄 Allow Intercom banners to be visible

nwingt:


williamchong (approved):
We should remove these and refactor the whole intercom plugin later, since APP use native now

PR #1029, opened 2026-04-27
✨ Bridge Intercom calls to native SDK in-app

williamchong:
Adds useIntercom() — a single composable that picks among the native Intercom bridge (in-app + supported), window.Intercom (browser), and mailto fallback. Returns { method } so analytics events can still report 'chat' vs 'link'. All six direct window.Intercom call sites (AppFooter, PricingPlanBenefits, use-claimable-books, account, store, reader/epub) now go through it.

Capability is detected via window.__nativeBridge.features injected by the native shell, so older app builds fall back to mailto without a sync release. Extends the identifyUser bridge payload with intercomToken / likerId / evmWallet / likeWallet for the native SDK's Identity Verification flow, and skips window.Intercom('shutdown'| 'update'|'trackEvent') in-app — native owns that surface. Defangs window.Intercom to a no-op in-app so any stray third-party call can't pop a duplicate web messenger over the native one.

Hoists nativeFeatureSupported() into utils/native-bridge.ts and declares the global window.__nativeBridge type once.

### DIFF
--- a/composables/use-intercom.ts
+++ b/composables/use-intercom.ts
@@ -1,0 +1,64 @@
+// Single entry point for Intercom on web. Picks among the native RN SDK
+// bridge, the window.Intercom JS SDK, and a mailto fallback so call sites
+// don't have to branch on isApp / SDK availability themselves.
+
+const SUPPORT_EMAIL = 'cs@3ook.com'
+
+type OpenResult = { method: 'chat' | 'link' }
+
+function openMailto(subject?: string, body?: string): void {
+  let mailto = `mailto:${SUPPORT_EMAIL}`
+  const params: string[] = []
+  if (subject) params.push(`subject=${encodeURIComponent(subject)}`)
+  if (body) params.push(`body=${encodeURIComponent(body)}`)
+  if (params.length) mailto += `?${params.join('&')}`
+  window.open(mailto, '_blank')
+}
+
+export function useIntercom() {
+  const { isApp } = useAppDetection()
+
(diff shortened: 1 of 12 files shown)


composables/use-intercom.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (typeof window?.Intercom === 'function') {
+      window.Intercom('showNewMessage', message ?? '')
+      return { method: 'chat' }
+    }
+    openMailto(mailtoSubject ?? message, message)
+    return { method: 'link' }
+  }
+
+  function trackEvent(name: string, params?: Record<string, unknown>): void {
+    if (isApp.value && isNativeFeatureSupported('intercom')) {
```
nwingt: can refactor `isNativeFeatureSupported('intercom')` to a util func?

composables/use-intercom.ts
```diff
@@ -0,0 +1,64 @@
...
+  function showNewMessage(message?: string, mailtoSubject?: string): OpenResult {
+    if (isApp.value) {
+      if (isNativeFeatureSupported('intercom')) {
+        postToNative({ type: 'intercomShowNewMessage', message })
+        return { method: 'chat' }
+      }
+      openMailto(mailtoSubject ?? message, message)
+      return { method: 'link' }
+    }
+    if (typeof window?.Intercom === 'function') {
```
nwingt: may refactor `typeof window?.Intercom === 'function'` to a util func

nwingt (approved):

PR #1050, opened 2026-04-29
[TAS-5398] ✨ Let authors manage Stripe Connect from account page

williamchong:
image
Adds a payout-account row that drives Stripe Express onboarding, resumes incomplete setup, or opens the dashboard when ready, and handles the post-onboarding return + link-refresh redirects.


i18n/locales/en.json
```diff
@@ -108,6 +110,7 @@
   "account_page_migrate_legacy_book": "Upgrade legacy books",
   "account_page_payment_currency": "Payment Currency",
   "account_page_payment_currency_button": "Choose currency",
+  "account_page_plus_title": "3ook Plus",
```
nwingt:
```suggestion
  "account_page_plus_title": "3ook.com Plus",
```

i18n/locales/zh-Hant.json
```diff
@@ -108,6 +110,7 @@
   "account_page_migrate_legacy_book": "升級舊版書籍",
   "account_page_payment_currency": "付款貨幣",
   "account_page_payment_currency_button": "選擇貨幣",
+  "account_page_plus_title": "3ook Plus",
```
nwingt:
```suggestion
  "account_page_plus_title": "3ook.com Plus",
```

pages/account/index.vue
```diff
@@ -674,6 +738,86 @@ const colorModeLabel = computed(
...
+  },
+  pending: {
+    status: 'account_page_stripe_connect_status_pending',
+    button: 'account_page_stripe_connect_resume_button',
+  },
+  none: {
+    status: 'account_page_stripe_connect_status_none',
+    button: 'account_page_stripe_connect_setup_button',
+  },
+}
```
nwingt: can we use computed with $t here for i18n-ally extension

nwingt (commented):
It seems the connected status remains when switching account without a refresh

Also, please review the [patch](https://github.com/williamchong/liker-land-v3/pull/24)

PR #1055, opened 2026-04-30
♻️ Use server timestamp for book setting markers

williamchong:
Replace client `Date.now()` with `FieldValue.serverTimestamp()` for `completedAt`, `didNotFinishAt`, `archivedAt`, and `lastOpenedTime` so stored times can't be poisoned by device clock skew. Wire payload now sends a `true` sentinel; the store keeps the client time in local cache so shelf sort updates immediately.


server/utils/api-user.ts
```diff
@@ -89,38 +90,34 @@ export async function getBookSettings(
...
-        : Timestamp.fromMillis(didNotFinishAt),
-    }),
-    ...(archivedAt !== undefined && {
-      archivedAt: archivedAt === null
-        ? FieldValue.delete()
-        : Timestamp.fromMillis(archivedAt),
-    }),
+    ...markerWrites,
+    ...(lastOpenedTime !== undefined && { lastOpenedTime: Date.now() }),
     updatedAt: FieldValue.serverTimestamp(),
```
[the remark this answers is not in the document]
nwingt: @williamchong seems true?

shared/types/book-settings.d.ts
```diff
@@ -14,3 +14,12 @@ export interface BaseBookSettingsData {
   'didNotFinishAt'?: number | null
   'archivedAt'?: number | null
 }
+
+export type ServerTimestampField = 'lastOpenedTime' | 'completedAt' | 'didNotFinishAt' | 'archivedAt'
```
nwingt: The name is too generic
```suggestion
export type BookTimestampField = 'lastOpenedTime' | 'completedAt' | 'didNotFinishAt' | 'archivedAt'
```

server/utils/api-user.ts
```diff
@@ -77,50 +104,39 @@ export async function getBookSettings(
...
   const {
-    updatedAt: _,
     completedAt,
     didNotFinishAt,
     archivedAt,
+    lastOpenedTime,
     ...restSettings
   } = settings
 
+  const markerWrites = Object.fromEntries(
```
nwingt:
```suggestion
  const timestampFields = Object.fromEntries(
```

nwingt (approved):

PR #1060, opened 2026-05-04
🚸 Show different modal for confirm delete

nwingt:
Before

After

### DIFF
--- a/i18n/locales/zh-Hant.json
+++ b/i18n/locales/zh-Hant.json
@@ -338,6 +338,8 @@
   "color_mode_system": "跟隨系統",
   "common_back": "返回",
   "common_cancel": "取消",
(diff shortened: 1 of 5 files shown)


i18n/locales/zh-Hant.json
```diff
@@ -338,6 +338,8 @@
   "color_mode_system": "跟隨系統",
   "common_back": "返回",
   "common_cancel": "取消",
+  "common_close": "關閉",
+  "common_delete": "删除",
```
williamchong:
```suggestion
  "common_delete": "刪除",
```

williamchong (changes requested):

williamchong (approved):

PR #1064, opened 2026-05-04
🐛 Fix TTS sample experiment exposure on pricing page

williamchong:


plugins/posthog-os-bootstrap.client.ts
```diff
@@ -0,0 +1,14 @@
+// Seed $initial_os into PostHog flag-eval person properties before the next
+// /decide call. With `person_profiles: 'identified_only'`, anonymous visitors
```
nwingt, line 2: is that an extra `/`?

williamchong: `/decide` should be an internal API or sth found by claude inside posthog lib

PR #1074, opened 2026-05-06
[TAS-5618] ✨ Support bookmark in reader

nwingt:
Based on #1051


server/utils/annotations.ts
```diff
@@ -17,14 +22,20 @@ function getAnnotationsCollection(userWallet: string, nftClassId: string) {
...
-    note: note || '',
-    chapterTitle: chapterTitle || '',
-    createdAt: createdAt instanceof Timestamp ? createdAt.toMillis() : Date.now(),
-    updatedAt: updatedAt instanceof Timestamp ? updatedAt.toMillis() : Date.now(),
+    type,
+    ...(data.cfi !== undefined ? { cfi: data.cfi } : {}),
+    ...(data.page !== undefined ? { page: data.page } : {}),
+    ...(data.text !== undefined ? { text: data.text } : {}),
+    ...(data.color !== undefined ? { color: data.color } : {}),
+    ...(data.note !== undefined ? { note: data.note } : {}),
```
williamchong, line 39: We should wrap this pattern in a later refactor

pages/reader/epub.vue
```diff
@@ -1930,6 +1969,88 @@ async function handleAnnotationNavigate(annotation: Annotation) {
   }
 }
 
+const currentBookmarkCfi = computed(() => currentPageStartCfi.value || currentCfi.value)
+const currentBookmark = computed(() => {
+  return currentBookmarkCfi.value ? getBookmarkByCfi(currentBookmarkCfi.value) : undefined
+})
+const isCurrentPageBookmarked = computed(() => !!currentBookmark.value)
```
williamchong, line 1975: Not sure if we should do a cfi range check from page start to end, instead of direct comparision

nwingt: Updated

williamchong (approved):

pages/reader/epub.vue
williamchong: the bookmark button now has `:disabled="isReaderLoading || !currentBookmarkCfi"` , which seems to be not wrong but might needs to be reviewed

pages/reader/epub.vue
```diff
@@ -1930,6 +1968,88 @@ async function handleAnnotationNavigate(annotation: Annotation) {
...
+    const cfi = new EpubCFI(bookmark.cfi)
+    if (cfi.range) {
+      cfi.collapse(true)
+    }
+    await rendition.value?.display(cfi.toString())
+  }
+  catch (error) {
+    console.error('Failed to navigate to bookmark:', error)
+    toast.add({
+      title: $t('reader_annotations_navigate_failed'),
```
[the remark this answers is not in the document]
nwingt: ignore for now

shared/schemas/annotation.ts
```diff
@@ -7,7 +7,8 @@ const AnnotationColorSchema = v.picklist(
   'MISSING_OR_INVALID_COLOR',
 )
 
-export const AnnotationCreateSchema = v.object({
+const HighlightCreateSchema = v.object({
+  type: v.literal('highlight', 'MISSING_OR_INVALID_TYPE'),
   cfi: v.pipe(
```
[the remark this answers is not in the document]
nwingt: client & server will be updated at the same time

williamchong (approved):

PR #1092, opened 2026-05-09
🔍️ Give each crawlable page a unique description

williamchong:


pages/store/index.vue
```diff
@@ -470,8 +470,32 @@ const tagName = computed(() => {
...
+      const displayName = ownerWalletDisplayName.value || queryOwnerWallet.value
+      return $t('store_page_owner_description', { owner: displayName })
+    }
+    if (queryGenre.value) {
+      return $t('store_page_genre_description', { genre: localizedGenreName.value })
+    }
+  }
+  const cmsDescription = isStakingTagId.value
+    ? ''
+    : activeCMSTag.value?.description[normalizedLocale.value] || ''
```
nwingt: Can we avoid the one-liner conditional for isStakingTagId here? Since this computed already has several explicit branches, a regular if would be easier to scan than the ternary

PR #1109, opened 2026-05-12
🐛 Dedupe PostHog purchase/trial/subscribe events via shared uuid

williamchong:
posthog-js 1.373.2 added a caller-supplied `uuid` to `CaptureOptions` (PostHog/posthog-js#3546), the feature c5d3e80 was waiting on. For the 4 events the API server also fires from the Stripe webhook handler, both sides now build the same URL
(`https://3ook.com/posthog-dedup/${eventName}/${transaction_id}`) and hash it under RFC 4122 NAMESPACE_URL to get a deterministic uuidv5, so PostHog's ClickHouse dedup tuple (timestamp, distinct_id, event, uuid) collapses the pair into one row. Removes the c5d3e80 workaround that suppressed the browser leg entirely.

The matching server change lives in likecoin-api-public/src/util/ posthog.ts — the URL string is the contract; both sides must build it identically (post-mapping event name, same path layout) or dedup breaks silently.

Requires https://github.com/likecoin/likecoin-api-public/pull/1253


composables/use-logger.ts
```diff
@@ -82,21 +83,28 @@ const INTERCOM_EVENT_ALLOWLIST = new Set<string>([
...
+const POSTHOG_SERVER_MIRRORED_EVENTS = new Set<string>([
   'begin_checkout',
   'purchase',
   'start_trial',
   'subscribe',
 ])
 
+// The host must be a fixed literal: substituting useRuntimeConfig().public.siteUrl or
+// any env-aware hostname would flip between testnet/mainnet and silently break dedup.
+function derivePostHogEventUuid(eventName: string, transactionId: string): string {
```
nwingt:
```suggestion
function derivePostHogEventUUID(eventName: string, transactionId: string): string {
```

composables/use-logger.ts
```diff
@@ -180,7 +186,13 @@ export function useLogEvent(eventName: string, eventParams: EventParams = {}) {
         posthogParams.nft_class_ids = classIds.join(',')
       }
     }
-    proxy.posthog.capture(eventName, { app: '3ook', ...posthogParams })
+    const transactionId = typeof posthogParams.transaction_id === 'string'
+      ? posthogParams.transaction_id
+      : undefined
+    const captureOptions = POSTHOG_SERVER_MIRRORED_EVENTS.has(eventName) && transactionId
+      ? { uuid: derivePostHogEventUuid(eventName, transactionId) }
```
nwingt:
```suggestion
      ? { uuid: derivePostHogEventUUID(eventName, transactionId) }
```

PR #1110, opened 2026-05-12
✨ Honor affiliate Plus-discount opt-in on store product page

williamchong:
Plumb the new `isPlusDiscountAllowed` flag from the affiliate API through to the store product page so Plus members see the 20% discount when an opted-in affiliate's `?from=@likerId` link is used. The tipping price- scaling now also gates on whether the discount will actually apply, to avoid inflating `customPrice` when the server won't discount it.

Extracted `useLinkAffiliate` to mirror `use-plus-affiliate` and `use-book-owner-affiliate`, gated on `isLikerPlus` to skip the fetch for non-Plus visitors on this high-traffic page.

Requires: https://github.com/likecoin/likecoin-api-public/pull/1254


AuroraHuang22 (approved):

composables/use-link-affiliate.ts
```diff
@@ -0,0 +1,44 @@
+import type { AffiliatePublicConfig } from '~/shared/types/affiliate'
+import { normalizeLikerId } from '~/shared/utils/liker-id'
+
+export function useLinkAffiliate(from: MaybeRefOrGetter<string | undefined>) {
```
nwingt: it seems this composable has nothing to do with link/route
```suggestion
export function useAffiliateConfig(affilateId: MaybeRefOrGetter<string | undefined>) {
```

composables/use-link-affiliate.ts
```diff
@@ -0,0 +1,44 @@
...
+  const { user } = useUserSession()
+
+  const loadedConfig = useState<AffiliatePublicConfig | null>('link-affiliate-config', () => null)
+  const loadedFrom = useState<string | null>('link-affiliate-loaded-from', () => null)
+  const isLoading = useState<boolean>('link-affiliate-loading', () => false)
```
nwingt:
```suggestion
  const loadedConfig = useState<AffiliatePublicConfig | null>('affiliate-config', () => null)
  const loadedFrom = useState<string | null>('affiliate-config-loaded-from', () => null)
  const isLoading = useState<boolean>('affiliate-config-loading', () => false)
```

composables/use-link-affiliate.ts
```diff
@@ -0,0 +1,44 @@
...
+    isLoading.value = true
+    try {
+      const likerId = normalizeLikerId(fromValue)
+      loadedConfig.value = await $fetch<AffiliatePublicConfig>(
+        `/api/affiliate/${encodeURIComponent(likerId)}`,
+      )
+      loadedFrom.value = fromValue
+    }
+    catch (error) {
+      console.error('[LinkAffiliate] Failed to fetch:', error)
```
nwingt:
```suggestion
      console.error('[AffiliateConfig] Failed to fetch:', error)
```

composables/use-link-affiliate.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function useLinkAffiliate(from: MaybeRefOrGetter<string | undefined>) {
+  const { user } = useUserSession()
+
+  const loadedConfig = useState<AffiliatePublicConfig | null>('link-affiliate-config', () => null)
+  const loadedFrom = useState<string | null>('link-affiliate-loaded-from', () => null)
+  const isLoading = useState<boolean>('link-affiliate-loading', () => false)
+
+  async function fetchConfig() {
+    const fromValue = toValue(from)
```
nwingt:
```suggestion
    const fromValue = toValue(affilateId)
```

composables/use-affiliate-config.ts
```diff
@@ -0,0 +1,60 @@
+import type { AffiliatePublicConfig } from '~/shared/types/affiliate'
+import { normalizeLikerId } from '~/shared/utils/liker-id'
+
+export function useAffiliateConfig(affiliateId: MaybeRefOrGetter<string | undefined>) {
+  const { user } = useUserSession()
+
+  const loadedConfig = useState<AffiliatePublicConfig | null>('affiliate-config', () => null)
+  const loadedFrom = useState<string | null>('affiliate-config-loaded-from', () => null)
+  const isLoading = useState<boolean>('affiliate-config-loading', () => false)
```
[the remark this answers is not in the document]
williamchong: there can only be 1 from qs in 1 session

PR #1124, opened 2026-05-14
⬆️ Upgrade minimax-speech-ts to v0.3.0

williamchong:
v0.3.0 drops the 'neutral' emotion (replaced with 'calm') and reshapes synthesizeStream's return to { audio, subtitle }; destructure audio at the call site. Default apiHost also moves to api.minimax.io.


williamchong:
@prayrequest

nwingt (approved):

PR #1136, opened 2026-05-15
🚀 Cache NFT metadata server-side to collapse blocking SSR fetch

williamchong:
Book/store SSR fetched aggregated NFT metadata fresh on every render, making the upstream LikeCoin call a CPU-bound stall on the dominant traffic class. Route the SSR fetch through /api/nft/:id/metadata, which wraps the upstream call in a short shared Nitro storage cache.

The payload is book-level (not session-scoped), so a shared cache is safe and per-user rendering is unaffected. The client keeps hitting LikeCoin directly — that endpoint is already Cloudflare-cached, so an extra hop through our server would only add latency. callOnce render mode + Pinia payload transfer mean no duplicate fetch or hydration mismatch from the server/client split.

### DIFF
--- a/server/api/nft/[nftClassId]/metadata.get.ts
+++ b/server/api/nft/[nftClassId]/metadata.get.ts
@@ -1,0 +1,50 @@
+import { FetchError } from 'ofetch'
+
+import type { LikeCoinNFTClassAggregatedMetadataOptionKey } from '~~/shared/utils/api'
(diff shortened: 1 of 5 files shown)


nwingt (approved):

nwingt:
Should we do the same to staked books in /store too?

williamchong:
> Should we do the same to staked books in /store too?

they should use the same API path

PR #1142, opened 2026-05-18
⬆️ Upgrade @nuxt/ui to v4

williamchong:
Bump @nuxt/ui 3.3.7 → 4.7.1 (Nuxt 4 already in place) and add tailwindcss as an explicit dependency, now required as a peer.

Rename the three UButtonGroup usages to UFieldGroup per the v4 breaking change (PDF reader page-nav/zoom controls, store sticky edition selector). No other v4 breaking changes apply — Marquee, Accordion, model modifiers, Form and the AI chat components are unused. Typecheck and lint pass.


nwingt (approved):
No issue found on sepolia so far

PR #1153, opened 2026-05-19
⬆️ Upgrade @likecoin/epub-ts to 0.6.4

williamchong:
0.6.4's Mapping.page() detects a fast-path-induced start/end range inversion and recomputes via the DOM walk, fixing at the source the inverted-CFI-span behaviour worked around in 5e37366a and the TTS-start fixes. Lockfile-only — package.json's ^0.6.3 already permits it. 5e37366a's order-agnostic guard stays as a downstream seatbelt (library recovery is gated on _measurer). TTS suite 59/59, typecheck clean.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5422,9 +5422,9 @@
       "license": "MIT"
     },
     "node_modules/@likecoin/epub-ts": {
-      "version": "0.6.3",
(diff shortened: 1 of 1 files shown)


nwingt (approved):
No issue found on sepolia so far

PR #1154, opened 2026-05-19
💬 Add zh-HK jyutping overrides for common misread chars

williamchong:
Covers 語塞, 自傳, 彈結他, 兮, 柙, 玩伴, 命裡/裏, 膜拜, 弶 —
each scoped with the fullest contiguous target so the dict entry stays selective and the payload lean.


server/utils/tts-minimax.ts
nwingt, line 1: Should we move the rules to a JSON file so that its easier for @ckxpress & @edmondyu to submit PRs

nwingt (approved):

PR #1184, opened 2026-05-26
🐛 Stop stale service worker from blocking PWA autoUpdate

williamchong:
The CDN was edge-caching /sw.js for hours, so the browser kept re-fetching the old worker and autoUpdate never detected new builds. Serve it 'no-cache' so it revalidates (cheap 304 via ETag) and the CDN won't hold a stale copy.


nwingt (approved):

nwingt:
@prayrequest

PR #1207, opened 2026-06-01
[TAS-5814] 🏗️ Source bookstore CMS from backend API instead of Airtable

nwingt:
Requires https://github.com/likecoin/likecoin-api-public/pull/1284


server/api/store/products.get.ts
```diff
@@ -1,68 +1,11 @@
...
+// Auto-derived lists (latest / free / drm-free) live at /store/products/<type>.
+export default defineEventHandler((event) => {
+  const query = getQuery(event)
+  const tagRaw = Array.isArray(query.tag) ? query.tag[0] : query.tag
+  const tag = tagRaw ? String(tagRaw) : ''
+  if (!tag) {
+    throw createError({ statusCode: 400, statusMessage: 'TAG_REQUIRED' })
   }
+  return respondWithBookstoreListing(event, opts => fetchBookstoreCMSProductsByTagId(tag, opts))
 })
```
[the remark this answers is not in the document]
nwingt: we use Nuxt dude

server/utils/bookstore.ts
```diff
@@ -0,0 +1,178 @@
...
+  const cheapest = getCheapestListedPrice(book.prices)
+  return {
+    id: book.classId || book.id || '',
+    classId: book.classId,
+    title: book.name,
+    imageUrl: book.thumbnailUrl,
+    locales: book.inLanguage ? [book.inLanguage] : undefined,
+    isDRMFree: !book.hideDownload,
+    isAdultOnly: book.isAdultOnly || undefined,
+    minPrice: book.minPrice ?? cheapest?.price,
```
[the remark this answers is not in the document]
nwingt: refer to:
https://github.com/nwingt/3ook-com/blob/d22f130e58da57ccd82294aadc5a37daf67f988e/server/utils/bookstore.ts#L78

app/utils/api.ts
```diff
@@ -37,24 +38,25 @@ export function fetchStakingBookNFTs({
...
+      query: {
+        tag: isBuiltIn ? undefined : tagId,
+        offset,
+        limit,
+        ts,
+      },
     },
-  })
+  )
 }
```
williamchong: I think no reason the caller needs to know when to call `/store/products/${tag}` vs `/store/product/` with `query.tag` just because some tags are built-in
Maybe we should change API to unify the interface, and just call a same API here no matter built-in or dynamic tag

server/utils/bookstore.ts
```diff
@@ -0,0 +1,178 @@
...
+  if (offsetRaw !== undefined && offsetRaw !== '') {
+    const offsetNum = Number(offsetRaw)
+    if (!Number.isInteger(offsetNum) || offsetNum < 0) {
+      throw createError({ statusCode: 400, statusMessage: 'INVALID_OFFSET' })
+    }
+    nextKey = offsetNum === 0 ? undefined : String(offsetNum)
+  }
+  try {
+    const result = await fetcher({ pageSize, nextKey })
+    setHeader(event, 'cache-control', nextKey ? 'no-store' : 'public, max-age=60')
```
williamchong: Why `no-store`? Can we expect response with `key` to be stable now?

nwingt: some nextKeys are numeric index
but i think we can have the same cache control as they rarely change

app/utils/api.ts
nwingt: Unified

williamchong (approved):

server/utils/bookstore.ts
```diff
@@ -0,0 +1,178 @@
...
+  const cheapest = getCheapestListedPrice(book.prices)
+  return {
+    id: book.classId || book.id || '',
+    classId: book.classId,
+    title: book.name,
+    imageUrl: book.thumbnailUrl,
+    locales: book.inLanguage ? [book.inLanguage] : undefined,
+    isDRMFree: book.hideDownload === false,
+    isAdultOnly: book.isAdultOnly || undefined,
+    minPrice: book.minPrice ?? cheapest?.price,
```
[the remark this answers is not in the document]
nwingt: separate concept

PR #1218, opened 2026-06-02
🔋 Drop perpetual spin on the index redirect splash

williamchong:
The `/` page only exists to navigateTo store/shelf, but its loader ran an ungated animate-spin. Under Suspense the splash can linger if the redirect or destination stalls, leaving an infinite CSS transform that keeps WebContent at high energy. Keep the static icon, drop the spin.


nwingt (approved):
unbelievable

PR #1220, opened 2026-06-02
⏪️ Revert server-side book-file proxy to unblock large downloads

williamchong:
Firebase App Hosting (Cloud Run) buffers responses and caps them at 32 MiB, so streaming book binaries through /api/book-file truncated large files — e.g. a 47 MB EPUB stalled at 32 MiB (68%). ebook-cors serves the full file when fetched directly, so revert to the client-direct cross-origin fetch (token is still present in session.user.token).

Keeps the shared getSessionToken refactor in place. A Range-chunked proxy (each slice under the 32 MiB buffer cap) is the planned proper fix to restore same-origin/token-hiding without truncation.


nwingt (approved):
so sad

PR #1221, opened 2026-06-02
🍱 Update Plus Open Graph share image

williamchong:
image

@nwingt please think of a better og so that our fb ads is not screwed, while I use this one to buy some time


public/images/og/plus.jpg
nwingt, line 1: why 2x the size?

williamchong: resized

PR #1239, opened 2026-06-09
[TAS-5935] ✨ Filter library search and genre results

nwingt:
Resolve https://github.com/likecoin/3ook-com/pull/1232#discussion_r3378188639

### DIFF
--- a/app/pages/store/index.vue
+++ b/app/pages/store/index.vue
@@ -49,14 +49,12 @@
             />
           </div>
         </div>
-        <h1
-          v-else
-          class="px-3 py-1 text-theme-cyan font-medium bg-theme-black border-1 border-theme-black rounded-full"
-        >
-          <span v-if="querySearchTerm">{{ $t('store_search_prefix') }}{{ querySearchTerm }}</span>
-          <span v-else-if="queryAuthorName">{{ $t('store_author_prefix') }}{{ queryAuthorName }}</span>
-          <span v-else-if="queryPublisherName">{{ $t('store_publisher_prefix') }}{{ queryPublisherName }}</span>
-          <span v-else-if="queryGenre">{{ $t('store_genre_prefix') }}{{ localizedGenreName }}</span>
+        <h1 v-else-if="searchModeContext">
+          <PillButton
+            is-active
+            is-static
+            :label="`${searchModeContext.titlePrefix}${searchModeContext.label}`"
+          />
         </h1>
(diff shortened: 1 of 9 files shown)


app/stores/bookstore.ts
```diff
@@ -144,18 +147,19 @@ export const useBookstoreStore = defineStore('bookstore', () => {
 
   const bookstoreSearchResultsByQueryMap = ref<Record<string, BookstoreSearchResults>>({})
 
-  const getBookstoreSearchResultsByQuery = computed(() => (query: string) => {
+  const getBookstoreSearchResultsByQuery = computed(() => (query: string, isPlusReadingEnabled = false) => {
+    const queryKey = getLibraryScopedCacheKey(query, isPlusReadingEnabled)
```
williamchong: The name is a bit confusing, `isPlusReadingEnabled` could mean if this feature is enabled, then `getLibraryScopedCacheKey` sounds like it is always getting a key related to library, but in fact `getLibraryScopedCacheKey` return generic key when `isPlusReadingEnabled = false`, and `isPlusReadingEnabled` is actually `isLibrary`

williamchong (changes requested):
I think we should just use `Library` or `PlusReading` but not both, then just `isPlusReading`/`isLibrary` instead of `isPlusReadingEnabled`

app/stores/bookstore.ts
nwingt: can we say
`isPlusReadingEnabled` is for book
`isLibrary` is for book list

app/pages/store/index.vue
```diff
@@ -49,14 +49,12 @@
...
-          <span v-else-if="queryAuthorName">{{ $t('store_author_prefix') }}{{ queryAuthorName }}</span>
-          <span v-else-if="queryPublisherName">{{ $t('store_publisher_prefix') }}{{ queryPublisherName }}</span>
-          <span v-else-if="queryGenre">{{ $t('store_genre_prefix') }}{{ localizedGenreName }}</span>
+        <h1 v-else-if="searchModeContext">
+          <PillButton
+            is-active
+            is-static
+            :label="`${searchModeContext.titlePrefix}${searchModeContext.label}`"
+          />
         </h1>
```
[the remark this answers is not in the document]
nwingt: ignore for now

williamchong (approved):

PR #1247, opened 2026-06-10
✨ Show per-book read & TTS time in bookshelf menu

williamchong:
image

Surface the server-tracked totalReadingTimeMs / totalTTSListeningTimeMs (previously stripped before reaching the client) as a subtle caption at the foot of each book's action menu, on the user's own shelf only. Zero totals are hidden and durations render human-readable (e.g. "1h 30m") via a new locale-aware formatDuration util.


app/components/BookshelfItem.vue
```diff
@@ -195,6 +201,21 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  // Read/TTS time totals are only available for the logged-in user's own shelf,
+  // so the time header is gated on isOwnShelf. The TTS prop uses a lowercase
+  // "Tts" so its kebab-case binding round-trips (an uppercase "TTS" would not).
+  isOwnShelf: {
```
nwingt: maybe name this prop `isMyBookshelf` instead of creating new term?

app/components/BookshelfItem.vue
```diff
@@ -195,6 +201,21 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  // Read/TTS time totals are only available for the logged-in user's own shelf,
+  // so the time header is gated on isOwnShelf. The TTS prop uses a lowercase
+  // "Tts" so its kebab-case binding round-trips (an uppercase "TTS" would not).
```
nwingt: `// The TTS prop uses a lowercase "Tts" so its kebab-case binding round-trips (an uppercase "TTS" would not).`

should move this above `totalTtsListeningTimeMs`?

nwingt (approved):

PR #1257, opened 2026-06-12
✨ List eligible books in affiliate voice note

williamchong:
image

Make the "推薦獨家聲音僅適用於指定書目" footnote clickable to open a modal listing the designated books and eligible publishers for an affiliate's exclusive voices.


app/components/AffiliateBookListItem.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <NuxtLink
+    v-if="bookstoreInfo"
+    :to="productPageRoute"
```
nwingt, line 4: not sure whether we need to guard product page in app

nwingt (approved):

PR #1270, opened 2026-06-15
🐛 Fix library 熱門 listing showing wrong books

williamchong:


app/stores/nft.ts
```diff
@@ -4,7 +4,22 @@ export const useNFTStore = defineStore('nft', () => {
   const nftClassByIdMap = ref<Record<string, Partial<NFTClass>>>({})
   const messagesByNFTClassIdMap = ref<Record<string, NFTBuyerMessage[]>>({})
 
+  // Class IDs whose class + bookstore metadata was fetched live this session.
+  // Those caches are persisted to localStorage and hydrate stale; this set
+  // (never persisted) marks what's been revalidated since load and drives the
+  // stale-while-revalidate path in lazyFetch.
+  const revalidatedClassIds = ref<Set<string>>(new Set())
```
nwingt: follow nftClassId above
```suggestion
  const revalidatedNFTClassIds = ref<Set<string>>(new Set())
```

PR #1281, opened 2026-06-16
🐛 Polyfill Array/String .at() for old iOS Safari

williamchong:
iOS 15.3 (Safari <15.4) lacks Array/String.prototype.at, crashing /store with "e.split(Ud).at is not a function" from a bundled dep. Defined non-enumerable to match native semantics.

### DIFF
--- a/app/plugins/polyfill.ts
+++ b/app/plugins/polyfill.ts
@@ -4,6 +4,38 @@ if (!Object.hasOwn) {
   }
 }
 
+// Resolve a relative index per the `at` spec: truncate toward zero, NaN to 0,
+// negative counts from the end. Returns -1 when out of range.
+function resolveRelativeIndex(length: number, index: number): number {
+  const i = Math.trunc(Number(index)) || 0
+  const resolved = i < 0 ? length + i : i
+  return resolved >= 0 && resolved < length ? resolved : -1
+}
+
+if (!Array.prototype.at) {
+  Object.defineProperty(Array.prototype, 'at', {
+    value: function <T>(this: T[], index: number): T | undefined {
+      const i = resolveRelativeIndex(this.length, index)
+      return i < 0 ? undefined : this[i]
+    },
+    writable: true,
+    configurable: true,
+    enumerable: false,
+  })
+}
+
+if (!String.prototype.at) {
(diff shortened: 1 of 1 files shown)


nwingt (commented):
pass this to @copilot

williamchong:
> [app/plugins/polyfill.ts](/likecoin/3ook-com/pull/1281/files/b147c130b5ba972b57cec2c222bfe5be4184ec2c#diff-a9e24191789697ba510fda9855da626b4a4d210bb245eb1f2f2f3466893a8f6a)
> Comment on lines +29 to +32
>     value: function (this: string, index: number): string | undefined {
>       const i = resolveRelativeIndex(this.length, index)
>       return i < 0 ? undefined : this[i]
>     },
> ### Copilot commented 
> Copilot
> Copilot
> AI
> [on Jun 16, 2026](https://github.com/likecoin/3ook-com/pull/1281#discussion_r3420378415)
> * Medium
> 
> String.prototype.at should return a full surrogate pair for astral symbols when the index points at a leading surrogate. Using `this[i]` only returns a single UTF-16 code unit, which diverges from the native `String.prototype.at` semantics.
> 
> Suggested changeset 1 (1)
> app/plugins/polyfill.ts
> Original file line number	Diff line number	Diff line change
> `@@ -27,8 +27,17 @@`
> ` if (!String.prototype.at) {`
> ` Object.defineProperty(String.prototype, 'at', {`
> ` value: function (this: string, index: number): string | undefined {`
> `- const i = resolveRelativeIndex(this.length, index)`
> `- return i < 0 ? undefined : this[i]`
> `+ const str = String(this)`
> `+ const i = resolveRelativeIndex(str.length, index)`
> `+ if (i < 0) return undefined`
> `+`
> `+ const first = str.charCodeAt(i)`
> `+ if (first >= 0xD800 && first <= 0xDBFF && i + 1 < str.length) {`
> `+ const second = str.charCodeAt(i + 1)`
> `+ if (second >= 0xDC00 && second <= 0xDFFF) return str.slice(i, i + 2)`
> `+ }`
> `+`
> `+ return str.charAt(i)`
> ` },`
> ` writable: true,`
> ` configurable: true,`
> Copilot uses AI. Check for mistakes.

claude says no

PR #1282, opened 2026-06-16
✨ Forward non-library engagement with idempotent retry

williamchong:
forwardPlusReadingUsage now forwards owned / non-Plus reads (split out as nonLibrary* fields) for publisher engagement stats, not just rev-share-eligible library reads. Each call carries a unique id and enables retry — safe now that the API dedups on it.

Deploy AFTER the likecoin-api change that adds the id dedup; otherwise a retried <no response> against the old API would double-count.

Requires: https://github.com/likecoin/likecoin-api-public/pull/1326

### DIFF
--- a/server/utils/plus-revshare-client.ts
+++ b/server/utils/plus-revshare-client.ts
@@ -1,3 +1,6 @@
+import { randomUUID } from 'node:crypto'
+import { API_MAX_RETRIES } from '~~/shared/utils/fetch-retry'
+
 interface PlusReadingUsageInput {
   isPaidPlus: boolean
   isBorrowed: boolean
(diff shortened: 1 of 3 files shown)


server/utils/plus-revshare-client.ts
```diff
@@ -8,20 +11,31 @@ interface PlusReadingUsageInput {
...
-  // Reading and TTS both count only for borrowed (Plus-library) books, and only
-  // for paid (non-trial) Plus, since trial usage must never fund the pool.
-  if (!input.isPaidPlus) return
   const { readerWallet, classId } = input
-  const readingTimeMs = input.isBorrowed ? input.readingTimeMs : 0
-  const ttsTimeMs = input.isBorrowed ? input.ttsTimeMs : 0
-  if (readingTimeMs <= 0 && ttsTimeMs <= 0) return
+  // Rev-share funds only borrowed (Plus-library) reads by paid (non-trial) Plus; the
+  // rest is non-library engagement, reported to publishers but never funding the pool.
+  const revShareEligible = input.isPaidPlus && input.isBorrowed
```
nwingt:
```suggestion
  const isRevShareEligible = input.isPaidPlus && input.isBorrowed
```

nwingt (approved):
may update outdated comments

PR #1289, opened 2026-06-17
🍱 Update OG images and add for /store and /library

nwingt:


public/images/og/store.jpg
williamchong, line 1: cut 與刊物
then we can have 3 clean new line for each sentence ie
瀏覽…
配備…
掌握…

public/images/og/default.jpg
williamchong, line 1: we can add 可讀 可聽 可擁有 here, unlikely to change

williamchong (approved):

PR #1292, opened 2026-06-18
✨ Add new featured publishers to the about page

williamchong:
Add 重版文化, 科學月刊, 啓明出版, 大銀 and Penana Limited (routed by owner wallet) plus the 崧博出版 imprints 崧燁文化, 財經錢線, 沐燁文化 and 樂律文化. The imprints share one wallet, so they route by publisher tag to keep each tile on its own books. Resolve avatars by wallet for the wallet-based entries.


nwingt (approved):
so no avatar for 崧博出版?

williamchong:
> so no avatar for 崧博出版?

maybe we should just add 1 entry for 崧博出版

PR #1293, opened 2026-06-18
[TAS-5084] ✨ Add footnote return button to EPUB reader

williamchong:
Tapping a footnote link does a one-way epub.js jump with no history, so readers had no way back to where they were. Capture the origin page on in-book link clicks and surface a floating "Return to reading" button that jumps back; it auto-dismisses once the reader is back at the origin.

### DIFF
--- a/app/pages/reader/epub.vue
+++ b/app/pages/reader/epub.vue
@@ -321,10 +321,11 @@
         <Transition name="reader-load">
           <UButton
             v-if="footnoteReturnCfi && !isReaderLoading"
-            class="absolute bottom-6 left-1/2 -translate-x-1/2 shadow-lg z-10"
+            class="absolute bottom-4.5 left-1/2 -translate-x-1/2 shadow-lg z-10"
             icon="i-material-symbols-u-turn-left-rounded"
             color="neutral"
             variant="solid"
+            size="sm"
             :label="$t('reader_footnote_return_button')"
             @click="handleFootnoteReturn"
           />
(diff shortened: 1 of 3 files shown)


i18n/locales/zh-Hant.json
```diff
@@ -766,6 +766,7 @@
   "reader_display_orientation_vertical": "直排",
   "reader_display_restore_defaults_button": "恢復預設",
   "reader_epub_rendition_display_failed": "無法顯示該頁。",
+  "reader_footnote_return_button": "返回閱讀內容",
```
nwingt, line 770:
```suggestion
  "reader_footnote_return_button": "回到註腳",
```

i18n/locales/en.json
```diff
@@ -766,6 +766,7 @@
   "reader_display_orientation_vertical": "Vertical",
   "reader_display_restore_defaults_button": "Restore Defaults",
   "reader_epub_rendition_display_failed": "The page cannot be displayed.",
+  "reader_footnote_return_button": "Return to reading",
```
nwingt, line 770:
```suggestion
  "reader_footnote_return_button": "Back to footnote",
```

app/pages/reader/epub.vue
```diff
@@ -316,6 +316,18 @@
...
         />
+
+        <Transition name="reader-load">
+          <UButton
+            v-if="footnoteReturnCfi && !isReaderLoading"
+            class="absolute bottom-6 left-1/2 -translate-x-1/2 shadow-lg z-10"
+            icon="i-material-symbols-u-turn-left-rounded"
+            color="neutral"
+            variant="solid"
+            :label="$t('reader_footnote_return_button')"
```
nwingt:
```suggestion
            class="absolute bottom-4.5 left-1/2 -translate-x-1/2 shadow-lg z-10"
            icon="i-material-symbols-u-turn-left-rounded"
            color="neutral"
            variant="solid"
            size="sm"
            :label="$t('reader_footnote_return_button')"
```

Before


After

i18n/locales/zh-Hant.json
williamchong: but this button goes back to reading content from footnote

i18n/locales/en.json
williamchong: but this button goes back to reading content from footnote

nwingt (approved):

PR #1294, opened 2026-06-18
✨ Add Techritual coverage to the about page

williamchong:


app/assets/images/about/media/techritual.webp
nwingt, line 1: Please use their square icon

https://www.techritual.com/wp-content/uploads/2024/02/cropped-techritual_logo_Square-270x270.png

Image

nwingt (changes requested):

PR #1303, opened 2026-06-22
💄 Update tab bar icon & label

nwingt:
<br>


app/assets/images/nav/bookshelf-zh.svg
williamchong, line 1: I think is is nice

williamchong (approved):

PR #1314, opened 2026-06-23
[TAS-6105] ✨ Add store intro banner for organic visitors

williamchong:
image

image

Greet first-time organic/direct visitors on the bare /store landing with a dismissable welcome banner pointing to /about. Gate it on the absence of campaign/affiliate attribution params, and dismiss on tag click — mirroring the existing LibraryIntroBanner pattern.


nwingt:
https://github.com/williamchong/liker-land-v3/pull/32

PR #1344, opened 2026-07-06
🚸 Offer support contact and shelf return when a book fails to open

williamchong:
Replace the single-button reader load-error modal with two explicit actions: contact customer service (pre-filling book id, format, error, url, status and platform for troubleshooting) and back to bookshelf. Covers EPUB, PDF and PDF render failures.

Extract the shared Intercom show-message + visibility helper so the new composable and the annotation report-issue flow no longer duplicate it.


i18n/locales/en.json
```diff
@@ -839,6 +839,10 @@
   "reader_display_orientation_vertical": "Vertical",
   "reader_display_restore_defaults_button": "Restore Defaults",
   "reader_epub_rendition_display_failed": "The page cannot be displayed.",
+  "reader_error_back_to_shelf_button": "Back to bookshelf",
+  "reader_error_contact_cs_button": "Contact customer service",
```
nwingt, line 843: button name too long? i guess the error modal footer may overflow on mobile

app/components/ErrorModal.vue
```diff
@@ -86,4 +86,13 @@ const LEVEL_STYLES = {
 } as const
 
 const levelStyle = computed(() => LEVEL_STYLES[props.level ?? 'error'])
+
+// Stack the footer buttons on mobile once there are multiple actions, so long
+// labels (e.g. contact support + back to shelf) don't overflow a narrow modal.
```
nwingt: the comment is too specific for a generic component

app/components/ErrorModal.vue
```diff
@@ -86,4 +86,13 @@ const LEVEL_STYLES = {
...
+const modalUI = computed(() => ({
+  title: 'flex items-center gap-2',
+  ...((props.actions?.length ?? 0) > 1
+    ? { footer: 'flex flex-col-reverse gap-2 sm:flex-row sm:justify-end' }
```
nwingt: i was thinking shorten button label text or adding flex-wrap, also no need to set conditionally, CC is overdoing this

better add items-center to center buttons horizontally with flex-direction is column ish
also items-center for flex row is nobrainer

nwingt (approved):

PR #1346, opened 2026-07-06
♻️ Refactor contract address, route auth and valibot schema config

williamchong:


server/schemas/params.ts
```diff
@@ -17,6 +17,13 @@ export const BookIdParamsSchema = v.object({
   nftClassId: bookIdField,
 })
 
+export const LikerIdParamsSchema = v.object({
+  likerId: v.pipe(
+    v.string('MISSING_LIKER_ID'),
+    v.nonEmpty('MISSING_LIKER_ID'),
```
nwingt, line 23: why both missing?

williamchong: original was `!likerId`, so we expect to guard `not set` (non string) and `empty` with same message

PR #1358, opened 2026-07-08
[TAS-6117] ✨ Add live search suggestions to store search input

williamchong:
Show a debounced dropdown of matching books as the user types in the store/library search, selecting one opens that book directly. Reuses the existing /api/store/search endpoint; skips fetching mid-IME-composition for CJK input and guards against stale responses overwriting newer terms.

### DIFF
--- a/app/pages/store/index.vue
+++ b/app/pages/store/index.vue
@@ -192,7 +192,7 @@
             <ul
               v-if="shouldShowSearchSuggestions"
               data-search-suggestions
-              class="border-t border-default max-h-80 overflow-y-auto py-2"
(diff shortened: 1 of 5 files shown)


app/pages/store/index.vue
```diff
@@ -183,6 +188,39 @@
                 </template>
               </UInput>
             </form>
+
+            <ul
+              v-if="shouldShowSearchSuggestions"
+              data-search-suggestions
+              class="border-t border-default max-h-80 overflow-y-auto py-2"
```
nwingt: use viewport unit instead of fixed 80 for better responsive support?

williamchong: cc suggest `max-h-[min(20rem,50svh)]` does that make sense?

PR #1361, opened 2026-07-08
♻️ Split account page into shared components 

williamchong:
Base on https://github.com/likecoin/3ook-com/pull/1359


nwingt (approved):

app/composables/use-plus-management.ts
nwingt, line 1: should we rename the composable to `use-plus-user-managment` so all Plus related files start with `use-plus*`?
Image

PR #1370, opened 2026-07-10
[TAS-6141] ✨ Let subscribers enter a referrer ID on /member

williamchong:
image

App Store installs cannot carry a `?from=` affiliate link the way Play's install referrer can, so IAP purchases lost their referrer attribution — exactly the path we want to grow, since IAP converts better than Stripe.

Add an optional referral code field, shown only when no `?from=` is already set. Applying a code writes `from=@<likerId>` into the route query, so a typed code and a clicked affiliate link converge on one code path: the alert, gift books, TTS voices, Stripe `from` and the RevenueCat `plusFrom` attribute all read it from there. No native shell changes needed.

Validation resolves the public Liker profile — there is no separate code registry, and a length check would reject legacy 7-20 char IDs. Only a 404 means "not found"; a network or 5xx failure must not tell a partner's customer their valid code does not exist.


i18n/locales/en.json
```diff
@@ -735,6 +735,14 @@
   "pricing_page_promo_first_monthly_free": "First month free",
   "pricing_page_promo_first_yearly_amount": "First year {currency} {price}",
   "pricing_page_promo_first_yearly_free": "First year free",
+  "pricing_page_referrer_input_confirm": "Apply",
+  "pricing_page_referrer_input_error_not_found": "We couldn't find this referral code. Please check and try again.",
+  "pricing_page_referrer_input_error_self": "You can't refer yourself.",
+  "pricing_page_referrer_input_error_unavailable": "Couldn't verify the code right now. Please try again.",
+  "pricing_page_referrer_input_help": "Applying a code may change the offer shown above.",
+  "pricing_page_referrer_input_label": "Referral code",
```
nwingt, line 743:
```suggestion
  "pricing_page_referrer_input_label": "Referral Code",
```

nwingt (commented):
https://github.com/williamchong/liker-land-v3/pull/33

ckxpress:
actually i've added a WIP in the notion task, thinking i'd discuss with you first (worry about reducing conversion. but i guess it's fine?)

is it only affecting web not app?

williamchong:
> actually i've added a WIP in the notion task, thinking i'd discuss with you first (worry about reducing conversion. but i guess it's fine?)
> 
> is it only affecting web not app?

It shows an additional clickable line in both version as long as a from qs is not currently set

PR #1384, opened 2026-07-13
[TAS-6162] ✨ Ask for a store rating once Plus is backend-confirmed

williamchong:
Sends `requestStoreReview` to the native shell after pollSessionUntilPlus() confirms the backend granted Plus, not when the store's purchase call returns. Entitlement is flipped asynchronously by the RevenueCat→backend webhook, so prompting on the purchase result would catch the user while the app still shows them as a non-subscriber — the surest way to earn a one-star review.

pollSessionUntilPlus() already returned this boolean and it was being discarded; navigation is unaffected, a slow webhook just means no prompt. The native shell applies its own engagement gate and both stores quota the prompt, so this is only ever a hint.

### DIFF
--- a/app/composables/use-subscription-checkout.ts
+++ b/app/composables/use-subscription-checkout.ts
@@ -194,7 +194,11 @@ export function useSubscriptionCheckout() {
           })
         }
         blockingModal.open({ title: $t('common_processing') })
-        await pollSessionUntilPlus()
+        const isPlusConfirmed = await pollSessionUntilPlus()
+        // Only once the backend has actually granted Plus: asking on the store's
+        // purchase result would catch the user while the app still shows them as a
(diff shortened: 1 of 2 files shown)


nwingt (approved):

ckxpress:
not sure about the UX. ideally, should ask whether the user is satisfied on a 5-star rating, then for 4 or 5 star responds, forward to app store/google play. for 1-3 star respond, forward to CS

williamchong:
> not sure about the UX. ideally, should ask whether the user is satisfied on a 5-star rating, then for 4 or 5 star responds, forward to app store/google play. for 1-3 star respond, forward to CS

But the native store popup is best effort and might not show up every time we call it

PR #1385, opened 2026-07-13
🐛 Show the 熱門 pill on the library's default tab

williamchong:
The library lands on the built-in `popular` listing, but unlike the other built-in list types it has no Airtable CMS tag mirroring it, and the pill row is built from CMS tags alone. So the active tab rendered no pill: the row silently started at 最新, and once a reader left 熱門 there was no way back to it.

Add the pill client-side, like the staking sort tabs, rather than relying on an editor creating the CMS record. It yields to a mirrored CMS tag if one is ever added, so editors keep control of the label and ordering.


williamchong:
added tag in CMS instead

